PR #2, opened 2021-09-21
Created Express.js Service

nathanielcaron:
- Created Express.js service
- Added version endpoint to prove functionality
- Added Winston logger (can now use logger.info('...') or logger.error('...') for example)

**IMPORTANT:**
You will need a `.env` file inside of the `/Service` folder to run the service, like this one:

https://drive.google.com/file/d/1a579vSNIiu3JVwUPrDpgL0SFjUXtSh7h/view?usp=sharing

These environment files are included in the `.gitignore` and should never be pushed to the repo since they will hold sensitive info like passwords for accounts and databases.


fredy20501:
Why is there a `package-lock.json` file outside of the Service folder? Shouldn't there be only 1 (i.e. the one in the Service folder)?

fredy20501 (approved):

sahilsaini432 (approved):

sahilsaini432:
@nathanielcaron I think there is a merge conflict with the README.md file.

PR #11, opened 2021-09-27
Mongoose DB setup

JoaoPontes98:
Installed mongoose and created the db connection. I added a config.js to act as a configuration for the DB and maybe other services. I wasn't really sure how to apply these credentials to environment variables. If you guys could let me know how this would be set up with environment variables that would be great because I'm a bit confused about it.

### DIFF
--- a/Service/app.js
+++ b/Service/app.js
@@ -1,9 +1,26 @@
 const EXPRESS = require('express');
 const APP = EXPRESS();
+const mongoose = require('mongoose');
+const dbConfig = require('./config').db;
 
 // Get environment
 require('dotenv').config();
 const ENV = process.env;
+
+//DB Credentials
+const username = dbConfig.username;
+const password = dbConfig.password;
+const dbName = dbConfig.name;
+const connectionString = `mongodb+srv://${username}:${password}@cluster0.pa1un.mongodb.net/${dbName}?retryWrites=true&w=majority`;
+mongoose.connect(connectionString);
+
+//Form DB connection
+const db = mongoose.connection;
+db.on('error', console.error.bind(console, 'connection error:'));
+db.once('open', function() {
+  // we're connected!
+  console.log("Mongodb connection successful");
+});
 
 // Define all routes in routes.js
 APP.use('/', require('./routes/routes'))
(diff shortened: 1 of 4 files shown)


nathanielcaron:
Ok so here is how to use environment variables with the dotenv package:

First, you'll need to add the variables the `.env` file:

db_name=Dev
db_user=dbAdmin
db_pass=<password>

Then inside of the file where you want to use the variables you'll need to import the package:
```
require('dotenv').config();
const ENV = process.env;
```

Then you can use ENV to get your variables:
e.g.,
const dbName = ENV.DB_NAME;

Please, don't merge this PR until the `config.js` file has been removed.
That way, no passwords are included in the code.


As a reminder `.env` files should never be pushed to the repo either.
Here is an example copy:
https://drive.google.com/file/d/1a579vSNIiu3JVwUPrDpgL0SFjUXtSh7h/view?usp=sharing

Service/app.js
```diff
@@ -1,10 +1,27 @@
...
+const dbName = dbConfig.name;
+const connectionString = `mongodb+srv://${username}:${password}@cluster0.pa1un.mongodb.net/${dbName}?retryWrites=true&w=majority`;
+mongoose.connect(connectionString);
+
+//Form DB connection
+const db = mongoose.connection;
+db.on('error', console.error.bind(console, 'connection error:'));
+db.once('open', function() {
+  // we're connected!
+  console.log("Mongodb connection successful");
```
nathanielcaron, line 22: We should always use the logger when printing information to the console instead of console.log:
```
logger.info('Mongodb connection successful');
```

JoaoPontes98:
Thanks for all the help @nathanielcaron!! I made the changes you suggested. Hope it's up to code B)

nathanielcaron:
> Thanks for all the help @nathanielcaron!! I made the changes you suggested. Hope it's up to code B)

Great work soldier! 💪

fredy20501 (approved):

fredy20501:
For whoever merges this PR, use **Squash** merge instead of normal merge so that the commit where the config.js file was pushed is not included in the merge. (if you don't know what I'm talking about, just don't merge it and someone will do it).

fredy20501:
Also might be a good idea to change the password of the DB user since it got pushed with the `config.js` file, even if we delete this branch afterwards.

JoaoPontes98:
> Also might be a good idea to change the password of the DB user since it got pushed with the `config.js` file, even if we delete this branch afterwards.

Sure! That's just a matter of changing it in MongoAtlas. I can do that when I get home.

PR #13, opened 2021-09-28
Add the dev process to the main README

fredy20501:
After our meeting discussion about the overall development process, I took notes and specified the details which are now in the README for all to see.

### DIFF
--- a/Documentation/MeetingNotes.md
+++ b/Documentation/MeetingNotes.md
@@ -5,6 +5,7 @@
   - [[Sept 21, 2021] First Official Meeting](#sept-21-2021-first-official-meeting)
   - [[Sept 23, 2021] Sprint 1 Planning](#sept-23-2021-sprint-1-planning)
   - [[Sept 28, 2021] Genesis (#1)](#sept-28-2021-genesis-1)
+  - [[Sept 30, 2021] Genesis (#1)](#sept-30-2021-genesis-1)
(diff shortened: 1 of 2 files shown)


sahilsaini432 (approved):

nathanielcaron (approved):

fredy20501:
Before we merge I think it's a good idea that everyone on the team reviews it so we are all on the same page for the dev process for now (it can always change later).

nathanielcaron:
Very detailed. I like it. 👍 
The only thing that is missing atm is mentioning to always run the tests and linting on the code before pushing changes.
We should never push code that does not follow our linting rules or doesn't pass unit tests
e.g., 
```
npm run test
npm run lint
git commit -m ""
...
```

JoaoPontes98 (approved):

fredy20501:
> Very detailed. I like it. 👍 The only thing that is missing atm is mentioning to always run the tests and linting on the code before pushing changes. We should never push code that does not follow our linting rules or doesn't pass unit tests e.g.,
> 
> ```
> npm run test
> npm run lint
> git commit -m ""
> ...
> ```

That's a good point, I'll add that to the document. 
Also later on (maybe next sprint?) I can create a quick GitHub Action so that linting & Jest unit tests are run when a PR is created (and we can set it so it needs to pass before it can be merged). I already have that set up in my last project so it would be a simple copy-paste.

nathanielcaron (approved):

PR #14, opened 2021-09-28
CBO-1 Create POST API Token endpoint

nathanielcaron:
Created new POST /auth endpoint.
Possible responses:
`200 OK, body: {  "token": "e361a0a7-33df-4c2b-816c-acfe9ff65528", "timestamp": 1632854754318 }`

Added verifyAuthToken check.
IMPORTANT: this check must be included in all routes of the API.
Possible responses:
`401 UNAUTORIZED, body: { 'No Authentication Token Provided' }` (when there is no `token` header)
`401 UNAUTORIZED, body: { 'Invalid Authentication Token Provided' }` (when `token` header is stale or invalid)
`200 OK, body: {endpoint specific response}`

Added 4 new unit tests for the `removeStaleTokens` and `isAuthTokenStale` util functions.

Updated README.

***I will add Karate API tests in a separate ticket and pull request.


Service/utils/utils.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+        expect(global.auth_tokens.has('b')).toBe(false);
+        expect(global.auth_tokens.size).toBe(2);
+    });
+
+    test('map is unchanged when token not in the map', () => {
+        UTILS.removeStaleTokens('d');
+
+        expect(global.auth_tokens.size).toBe(2);
+    });
```
fredy20501, line 22: Small detail:
Here the tests are both trying to modify the same data which can allow them to influence each other. For example, if the first test fails to remove the token (for whatever reason) then the second one will also fail since the size would still be 3. So 1 issue would cause 2 test failures.

In cases like this when tests both modify the same data we should use `beforeEach` and `afterEach` to setup and teardown the data for each individual test.

Service/utils/utils.js
```diff
@@ -0,0 +1,16 @@
+exports.removeStaleTokens = (token) => {
+    // Clear the stale token
+    if (auth_tokens.has(token)) {
+        auth_tokens.delete(token);
+    }
+
+    // Never let auth token map grow larger than 50,000 entries
+    if (auth_tokens.size >= 50000) {
+        auth_tokens.clear();
+    }
```
fredy20501, line 10: This is minor but this case does not have a unit test. (I realized it when I ran the jest coverage tool - it would be nice to keep coverage at 100% 😜)

fredy20501 (commented):
Great work! 
There are a few minor details that might be good double-check (see my other comments). I tested it locally with postman to make sure I understood how it worked and everything went as expected.

Service/utils/utils.js
nathanielcaron: Updated

Service/utils/utils.test.js
nathanielcaron: Updated

fredy20501 (approved):
Well done! 👍

PR #15, opened 2021-09-28
added eslint rules for service

sahilsaini432:


fredy20501:
You can also add a command for linting in `package.json` under `"scripts"` so we can run it manually. Something like `"lint": "eslint **/*.js"` (I did the same when I added eslint to the frontend).

Also, I ran it on my machine and got linting errors because of the line breaks (I run windows so line breaks are CRLF but it wants LF which is the line break for Unix). We might want to add a rule to ignore line break style (like adding `"linebreak-style": [ "off" ]` to the rules in the config file).

sahilsaini432:
@fredy20501 i just changed the code to turn off `linebreak-style` and also added the command to manually run it. It should work well on windows now. Let me know if there are any more issues.

fredy20501 (approved):
All good!

nathanielcaron (approved):

PR #23, opened 2021-10-12
DBO-19 Update Schema and Add DB Unit Tests

JoaoPontes98:
I added a /Service/db to house our db files. In there is the updated schema along with a few unit tests for the db. I also added a dbLib.js file to handle the different connection strings and other potential db functions. I also refactored the db connection in app.js to use the dbLib.js file

There are also more unit tests to be made for the other collections. But I didn't make them since we still don't have a set-in-stone data model until after we have the final wireframing. I just thought to make the unit tests after that since I think it would be more work to make the tests now and then fix them later.

P.S. I had to disable a linting-check in db.test.js, because when checking for unique id's in mongoose, you do <Name>._id. The linter doesn't like it :/


Service/db/dbSchema.js
```diff
@@ -0,0 +1,38 @@
...
+const { Schema } = mongoose;
+
+const postSchema = new Schema({
+  date_created: { type: Date, default: Date.now },
+  body: { type: String },
+  tags: [{ type: String }],
+  title: { type: String },
+  imgURL: { type: String }
+});
+const Post = mongoose.model('Post', postSchema, 'Post');
```
fredy20501, line 11: You can remove this post schema since it is the same as the new message schema.

Service/db/db.test.js
```diff
@@ -0,0 +1,55 @@
...
+});
+
+describe('Ensuring Clean DB After Tests', () => {
+  test('Check User Collection', async () => {
+    // Get the whole user collection
+    const UserCollection = await User.find();
+
+    expect(UserCollection).toEqual([]);
+  });
+});
```
fredy20501, line 55: This test might lead to some intermittent failures. For example, if 2 people create a new PR, the unit tests for both PRs will run at the same time within our GitHub Actions which means one of them might be checking for an empty collection at the same time as the other is adding stuff to it.

Instead of this test we could just check once in a while whether there is some residual stuff in the test db and manually clear it (and also try to figure out why they didn't get deleted).

Service/db/db.test.js
```diff
@@ -0,0 +1,55 @@
...
+    // Update testUser2's email
+    await User.findByIdAndUpdate(testUser2._id, update);
+    const testUserCheck = await User.findById(testUser2._id);
+
+    expect(testUserCheck.email).toBe(update.email);
+
+    // Delete testUser2 from db
+    await User.findByIdAndDelete(testUser2._id);
+  });
+});
```
fredy20501, line 46: Since you are deleting the test data within the test, it means that if the test fails it will stop and not reach the delete command, leading to test data not being cleared.

Potential solution: Have a global list defined at the top of the file. Whenever a test creates a test user (immediately after the `.save()`), add their ID to the list. Then you can add a global `AfterEach` block where you remove things from the list and delete them. (The `AfterEach` block always runs even if the test fails so it would solve the issue). Note in the future there would have to be a list for each type of object (i.e., `User` & `Message`) since you need to know which collection to delete from.

Service/app.js
```diff
@@ -27,13 +26,14 @@ if (ENV.NODE_ENV === 'dev') {
 }
 
 // DB Connection
-const connectionString = `mongodb+srv://${ENV.DB_USER}:${ENV.DB_PASS}@cluster0.pa1un.mongodb.net/${ENV.DB_NAME}?retryWrites=true&w=majority`;
-mongoose.connect(connectionString);
-const db = mongoose.connection;
-db.on('error', function() {
+const mongoose = require('mongoose');
+const db = require('./db/dbLib');
+db.connectDev(); // change to db.connectProd() when appropriate
```
fredy20501, line 31: Here we should implement the `dev` vs `prod` logic by checking the `NODE_ENV` environment variable:
```js
if (ENV.NODE_ENV === 'production') {
  db.connectProd();
}
else {
  db.connectDev();
}
```
This will be a lot easier than manually changing this line of code when switching to production.

fredy20501 (changes requested):
See comments below.

Service/db/dbLib.js
```diff
@@ -0,0 +1,22 @@
+const mongoose = require('mongoose');
+require('dotenv').config();
+const ENV = process.env;
+const devConnectionString = `mongodb+srv://${ENV.DB_USER}:${ENV.DB_PASS}@cluster0.pa1un.mongodb.net/${ENV.DB_DEV_NAME}?retryWrites=true&w=majority`;
+const testConnectionString = `mongodb+srv://${ENV.DB_USER}:${ENV.DB_PASS}@cluster0.pa1un.mongodb.net/${ENV.DB_TEST_NAME}?retryWrites=true&w=majority`;
+const prodConnectionString = `mongodb+srv://${ENV.DB_USER}:${ENV.DB_PASS}@cluster0.pa1un.mongodb.net/${ENV.DB_PROD_NAME}?retryWrites=true&w=majority`;
```
fredy20501, line 6: I just tried to run it locally and found a new issue: when you connect with Mongoos to a database with a name that doesn't exist, it creates a new database with that name. For example, since I didn't have the `DB_TEST_NAME` it resolved to `undefined` which meant that it created a new database called `undefined` in our MongoDB cluster. To prevent this issue we might want to add an extra check to make sure the name isn't `undefined`. Although this isn't a big issue to be fair.

JoaoPontes98:
> Sorry, apparently I closed this PR by mistake. Don't know how that happened 🤷 😅

Hahaha np! And this is all amazing feedback, thanks! I'll get on the fixes later this week, depending on when I finish my OS assignment lol

fredy20501:
Dammit, I did it again 🙄 oof.
Anyway, I know it looks like a lot so please do take your time. No rush! 😄

Service/db/dbLib.js
```diff
@@ -0,0 +1,35 @@
+const mongoose = require('mongoose');
```
nathanielcaron, line 1: If I understand correctly, this file is a collection of useful util functions for the database?

Could we rename it to `dbUtils.js`?
We could also move it to the utils folder unless you prefer keeping it with the other database stuff.
I'm totally ok with that.

nathanielcaron:
Great work @JoaoPontes98

nathanielcaron (approved):

Service/db/dbUtils.js
```diff
@@ -0,0 +1,35 @@
...
+};
+
+module.exports.connectProd = async () => {
+  await mongoose.connect(prodConnectionString);
+};
+
+module.exports.closeDatabase = async () => {
+  await mongoose.connection.dropDatabase();
+  await mongoose.connection.close();
+};
```
fredy20501, line 23: This function is pretty dangerous since it drops the current database so we should probably add some safety feature to prevent accidentally deleting our dev or prod db. 
Here's what I suggest:
- In the `closeDatabase` function, check to make sure the current database name is not the same as the dev/prod databases. (I believe you can get the current db name using `mongoose.connection.name`). If it is the same as the dev/prod db than we should probably throw an error and/or log it as an error.
- Rename the function to `deleteDatabase()` to clarify what it really does.

JoaoPontes98: Yeah checking the names is a great idea for safety. I added the changes in the newest commit. The `deleteDatabase()` now checks for the dev and prod db names and throws an error if found. I tested the function and it successfully prevents deleting the prod/dev db's. I also added a `closeDatabase()` to allow closing dev/prod connections.

nathanielcaron: Very nice

PR #32, opened 2021-11-03
DBO-8 Added api endpoints for User Posts

JoaoPontes98:
This includes api.raml docs, the addition of middleware to app.js, and updates to routes.js. I also made some syntax changes in the api.js that the linter was yelling at me to fix.

As you can see, these endpoints only allow for getting/deleting 1 post at a time. I didn't add a getAllUserposts() endpoint since I wasn't sure if it was necessary or not. If you guys think it needs to be added, I can whip it up real quick.

If you have more suggestions for response codes to add, please let me know.

### DIFF
--- a/Service/api/api.js
+++ b/Service/api/api.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-underscore-dangle */
 const uuidv4 = require('uuid').v4;
 const UTILS = require('../utils/utils');
 const { UserPost } = require('../db/dbSchema');
@@ -12,6 +11,7 @@ exports.createAuthToken = (req, res) => {
   return res.send(response);
 };
 
+// eslint-disable-next-line consistent-return
 exports.verifyAuthToken = (req, res, next) => {
   const token = req.header('token');
   const currentTime = Date.now();
@@ -40,7 +40,7 @@ exports.version = (req, res, next) => {
 
 exports.createUserPost = (req, res) => {
   const newUserPost = new UserPost({
-    author: req.body.author,
+    author: req.body.authorID,
     body: req.body.body,
     tags: req.body.tags,
     title: req.body.title,
@@ -62,7 +62,7 @@ exports.createUserPost = (req, res) => {
     return res.status(201).json({
       post: {
         _id: newUserPost._id,
-        author: newUserPost.author,
+        author: newUserPost.authorID,
         body: newUserPost.body,
         tags: newUserPost.tags,
         title: newUserPost.title,
@@ -77,14 +77,18 @@ exports.createUserPost = (req, res) => {
 };
 
 exports.deleteUserPost = (req, res) => {
+  // Find the userpost with the matching ID
   UserPost.findById(req.params.id)
     .then((doc) => {
       if (!doc) {
         return res.status(404).send('User Post not found.');
       }
+
       if (!('access_key' in req.body)) {
         return res.status(400).send('Client body does NOT contain an access-key.');
       }
+
+      // Check if the req & userpost access_key's match for deletion
       if (doc.access_key === req.body.access_key) {
         UserPost.findByIdAndDelete(req.params.id)
           .catch((err) => {
@@ -93,7 +97,7 @@ exports.deleteUserPost = (req, res) => {
           });
         return res.status(200).end();
       }
-      return res.status(401).send('Client access-key does NOT match the user post.');
+      return res.status(403).send('Client access-key does NOT match the user post.');
     })
     .catch((err) => {
       logger.error(err);
--- a/Service/api/api.js
+++ b/Service/api/api.js
@@ -1,38 +1,116 @@
+/* eslint-disable no-underscore-dangle */
 const uuidv4 = require('uuid').v4;
 const UTILS = require('../utils/utils');
+const { UserPost } = require('../db/dbSchema');
 
 exports.createAuthToken = (req, res) => {
-    let uuid = uuidv4();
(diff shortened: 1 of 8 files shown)


nathanielcaron:
Awesome work @JoaoPontes98! 🚀

Service/api/api.js
```diff
@@ -1,38 +1,116 @@
+/* eslint-disable no-underscore-dangle */
```
fredy20501, line 1: We might as well disable this rule globally instead of doing it for each file.

Just add this to the `.eslintrc.json` file in the "rules" portion:
```
"no-underscore-dangle": [ "off" ]
```
(then you should be able to remove this comment and the linter should not care)

Service/api/api.js
```diff
@@ -1,38 +1,116 @@
...
 exports.version = (req, res, next) => {
-    const VERSION = process.env.VERSION;
-    return res.send(`Service v${VERSION}`);
+  const { VERSION } = process.env;
+  return res.send(`Service v${VERSION}`);
+};
+
+exports.createUserPost = (req, res) => {
+  const newUserPost = new UserPost({
+    author: req.body.author,
```
fredy20501, line 43: The front end will not give you an author object but just the name and avatar image url. You'll need to create a user first using that given name & avatar image and then use that to create the post.

Service/api/api.js
```diff
@@ -1,38 +1,116 @@
...
+    body: req.body.body,
+    tags: req.body.tags,
+    title: req.body.title,
+    imgURL: req.body.imgURL,
+    date_created: req.body.date_created,
+    location: req.body.location,
+    true_location: req.body.true_location,
+    access_key: req.body.access_key
```
fredy20501, line 51: The access key will not be given by the frontend, the backend should be the one generating it and making sure it is unique (can double-check that it is unique by searching the db to see if another post already has the same access key).
You have already done a lot of work so writing the function to generate & verify access-keys could be done in another ticket.

Service/api/api.raml
```diff
@@ -17,6 +17,36 @@ types:
...
+        type: string
+      tags:
+        type: array
+        items:
+          type: string
+      title:
+        type: string
+      imgURL:
+        type: string
+      data_created:
```
fredy20501, line 37: Little typo 😉 
```suggestion
      date_created:
```

fredy20501 (commented):
Nice work! 👍

Service/api/api.js
fredy20501: Like Nathaniel said during our meeting, creating the user will be done in a separate function so you don't have to worry about this for now.

Service/api/api.js
```diff
@@ -1,38 +1,116 @@
...
+      }
+      if (doc.access_key === req.body.access_key) {
+        UserPost.findByIdAndDelete(req.params.id)
+          .catch((err) => {
+            logger.error(err);
+            return res.status(500).send(err);
+          });
+        return res.status(200).end();
+      }
+      return res.status(401).send('Client access-key does NOT match the user post.');
```
fredy20501, line 96: We have to be careful returning status code 401 since that's the one used to signify an invalid auth token. You should consider changing this return code to something different or changing the status code for an invalid auth token so it is different.

sahilsaini432 (changes requested):

Service/api/api.js
```diff
@@ -1,38 +1,116 @@
...
+    })
+    .catch((err) => {
+      logger.error(err);
+      return res.status(500).send(err);
+    });
+};
+
+exports.getUserPost = (req, res) => {
+  UserPost.findById(req.params.id)
```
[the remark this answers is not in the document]
JoaoPontes98: Yeah it's for getting posts from the db.

Service/api/api.js
```diff
@@ -12,6 +11,7 @@ exports.createAuthToken = (req, res) => {
   return res.send(response);
 };
 
+// eslint-disable-next-line consistent-return
```
fredy20501, line 14: We shouldn't disable linting for individual files/functions since the whole point of a linter is for **all** of our code to be consistent and follow all of the same rules. If you really think that an eslint rule is not appropriate for us, disable it globally. Otherwise, just change your code to make it work.

In this case, you can just replace `next()` with `return next()` to satisfy the linter. 
(you can google the eslint error to understand what it is expecting: e.g., [eslint consistent-return](https://eslint.org/docs/rules/consistent-return))

Service/api/api.js
fredy20501: I added a ticket on trello for implementing this function.

Service/api/api.js
JoaoPontes98: Yeah, that was definitely a lazy approach of me going about it, my bad. I didn't realize `return next()` was possible.

Should be fixed now with the latest commit.

fredy20501 (approved):
Awesome! 💯

sahilsaini432 (approved):

sahilsaini432 (approved):
Nice Work 👍

fredy20501:
Manual testing of the API endpoints done with Postman. Steps:
1. Got token using /auth endpoint (used it for all requests).
2. Got a userPost using the /userpost/{id} GET endpoint (got the ID from MongoDB).
3. Created a userPost using the /userpost POST endpoint (fails when required fields missing, as expected).
4. Delete a userPost using the /userpost/{id} DELETE endpoint (fails when access-key not given in body or not match, as expected).

PR #33, opened 2021-11-04
DBO-3 Linking application and service

sahilsaini432:
1. Updated Welcome Page to generate the auth token when you enter the site.
2. Updated Homepage to make get version as soon as you enter the site.
3. Added all API calls to the 'api' folder.

image
Current Home Page

### DIFF
--- a/application/src/App.test.js
+++ b/application/src/App.test.js
@@ -51,18 +51,4 @@ describe('Welcome Page', () => {
     expect(checkBox).not.toBeChecked();
     expect(welcomeButton).toBeInTheDocument();
   });
-
-  test('when checkbox is selected and welcome button is pressed', () => {
-    // Select the checkBox    
-    userEvent.click(checkBox);
-    // Click the button
-    fireEvent.click(welcomeButton);
-    // After the button is pressed it should not be on the screen
-    expect(welcomeButton).not.toBeInTheDocument();
-
-    // Get homepage components from the screen
-    const homepage = screen.getByTestId('home-page');
-    // Check to see if the "Homepage" is on the screen
-    expect(homepage).toBeInTheDocument();
-  });
 })
--- a/application/src/api/api.js
+++ b/application/src/api/api.js
@@ -1,0 +1,49 @@
+/* eslint-disable quotes */
+/* eslint-disable no-unused-vars */
+import axios from 'axios';
+import logger from '../logger/logger';
+
+// Auth Token
+let authToken;
+
+// Get environment
+require('dotenv').config();
+const ENV = process.env;
+
+// Get the service url from the environment file
+const serviceUrl = ENV.REACT_APP_SERVICE_URL;
+
+// Get Auth token
+export const getAuthToken = async () => {
+  try {
+    const response = await axios({
+      method: 'POST',
+      url: serviceUrl+'/auth'
+    })
+    authToken = response.data.token;
+  } catch(error) {
+    logger.info(error);
+  }
+}
+
+// Get Versions
+export const getVersion = async () => {
+  try {
+    const response = await axios({
+      method: 'GET',
+      url: serviceUrl+'/version',
+      headers: {
+        token: authToken
+      }
+    });
+
+    return response.data;
+  } catch(error) {
+    console.log(error);
+    if(error.response.status === 401) {
+      // Get New Auth Token
+      await getAuthToken();
+      return null;
+    }
+  }
+}
--- a/application/src/api/api.js
+++ b/application/src/api/api.js
@@ -8,6 +8,9 @@ let authToken;
 require('dotenv').config();
 const ENV = process.env;
 
+// Retry Variables
+const MAX_RETRY_LIMIT = 3;
+
 // Get the service url from the environment file
 const serviceUrl = ENV.REACT_APP_SERVICE_URL;
 
@@ -26,23 +29,35 @@ export const getAuthToken = async () => {
 
 // Get Versions
 export const getVersion = async () => {
-  try {
-    const response = await axios({
-      method: 'GET',
-      url: serviceUrl+'/version',
-      headers: {
-        token: authToken
+  return await requestWithToken(async() => {
+      const response = await axios({
+          method: 'GET',
+          url: serviceUrl+'/version',
+          headers: {
+              token: authToken
+          }
+      });
+      return response.data;
+  });
+}
+
+const requestWithToken = async (request) => {
+  for (let i=0; i<MAX_RETRY_LIMIT; i++) {
+      try {
+          // Make the given request
+          return await request();
+      } catch(error) {
+          if(error?.response?.status === 401) {
+              // Get New Auth Token and retry
+              await getAuthToken();
+          }
+          else {
+              // Don't retry if a different error occurs
+              logger.warn(error);
+              return null;
+          }
       }
-    });
-
-    return response.data;
-  } catch(error) {
-    logger.warn(error);
-
-    if(error?.response?.status === 401) {
-      // Get New Auth Token
-      await getAuthToken();
-      return null;
-    }
   }
+  logger.warn('Unable to get auth token after '+MAX_RETRY+' tries.');
+  return null;
 }
--- a/application/src/api/api.js
+++ b/application/src/api/api.js
@@ -9,7 +9,7 @@ require('dotenv').config();
 const ENV = process.env;
 
 // Retry Variables
-const MAX_RETRY_LIMIT = 3;
+const MAX_RETRY_LIMIT = 2;
 
 // Get the service url from the environment file
 const serviceUrl = ENV.REACT_APP_SERVICE_URL;
--- a/application/src/api/api.js
+++ b/application/src/api/api.js
@@ -21,8 +21,9 @@ export const getAuthToken = async () => {
       url: serviceUrl+'/auth'
     })
     authToken = response.data.token;
+    console.log('Token Created: '+authToken);
   } catch(error) {
-    logger.info(error);
+    logger.warn(error);
   }
 }
 
@@ -39,7 +40,8 @@ export const getVersion = async () => {
 
     return response.data;
   } catch(error) {
-    console.log(error);
+    logger.warn(error);
+
     if(error.response.status === 401) {
       // Get New Auth Token
       await getAuthToken();
--- a/application/src/api/api.js
+++ b/application/src/api/api.js
@@ -21,7 +21,6 @@ export const getAuthToken = async () => {
       url: serviceUrl+'/auth'
     })
     authToken = response.data.token;
-    console.log('Token Created: '+authToken);
   } catch(error) {
     logger.warn(error);
   }
--- a/application/src/api/api.js
+++ b/application/src/api/api.js
@@ -58,6 +58,6 @@ const requestWithToken = async (request) => {
           }
       }
   }
-  logger.warn('Unable to get auth token after '+MAX_RETRY+' tries.');
+  logger.warn('Unable to get auth token after '+MAX_RETRY_LIMIT+' tries.');
   return null;
 }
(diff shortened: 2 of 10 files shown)


application/src/components/WelcomePage.js
```diff
@@ -8,14 +8,22 @@ import './components.css';
...
+  // Make auth token API Call
+  const getToken = async () => {
+    await getAuthToken();
+  }
+
   // Handler function for the welcome button
   const enterButtonHandler = () => {
     if (agree == true) {
+      // Generate New Authtoken
+      getToken();
```
nathanielcaron, line 26: I think you should make `enterButtonHandler` async and call the API function directly.
You should never call an async function (`getToken`) from a non-async function (`enterButtonHandler`).

application/src/api/api.js
```diff
@@ -0,0 +1,49 @@
...
+      method: 'GET',
+      url: serviceUrl+'/version',
+      headers: {
+        token: authToken
+      }
+    });
+
+    return response.data;
+  } catch(error) {
+    console.log(error);
```
nathanielcaron, line 42: No `console.log` statements. Use the logger instead.

application/src/api/api.js
```diff
@@ -0,0 +1,49 @@
...
+// Get Auth token
+export const getAuthToken = async () => {
+  try {
+    const response = await axios({
+      method: 'POST',
+      url: serviceUrl+'/auth'
+    })
+    authToken = response.data.token;
+  } catch(error) {
+    logger.info(error);
```
nathanielcaron, line 25: Should use logger.warn instead since its an error

nathanielcaron:
Good stuff 💯

application/src/components/HomePage.js
```diff
@@ -1,13 +1,44 @@
...
+  useEffect(() => {
+    const fetchData = async () => {
+      // Get app version
+      let appVersion = await getVersion();
+      setVersion(appVersion);
+      setLoading(false);
+    }
+
+    fetchData();
+  }, [version, isLoading]);
```
fredy20501, line 27: Currently the `getVersion()` function returns null when the auth token was invalid. When that happens we want to try the request again. I did some testing and it seems like it tries 3 times if it fails, currently. I don't know enough about React and the `useEffect` and `useState` functions to know if that is intended or not.

Do you know what makes it try 3 times and whether or not this will be consistent?

sahilsaini432: Based on the documentation I read about `useEffect` its only called when we re-render the component or when the value of the dependencies changes. In this case the `useEffect` has two dependencies `version` and `isLoading`.

application/src/logger/logger.js
```diff
@@ -1,15 +1,8 @@
...
-        all: true
-    }),
-);
-
 const logger = winston.createLogger({
     level: 'debug',
-    format: winston.format.json(),
-    transports: [ new winston.transports.Console({ format: winston.format.combine(winston.format.colorize(),customFormat) }) ]
+    transports: [ new winston.transports.Console() ]
 });
```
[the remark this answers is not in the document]
sahilsaini432: It was giving out some random values in the console and also it was not adding any colour. So I just removed it.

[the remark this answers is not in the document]
fredy20501: Ah, so for me (using Chrome) the colour worked just fine in the console. I'm guessing you were using another browser that probably doesn't support colour in the console.

In any case, the colour wasn't necessary anyway so it's fine to remove it. I was just curious 😄

application/src/components/HomePage.js
fredy20501: Ah ok. So based on that the reason it is called 3 times is because it is called:
1. First when the component loads.
2. Again since you are changing the `version` using `setVersion()`.
3. A third time since you are changing the `isLoading` state using `setLoading()`.

I don't think that's how it should work. I don't think the `useEffect` should depend on `version` and `isLoading` since we don't want to call the function whenever those values change, but only when the component/page loads, right?
(Currently, it will always be triggered 3 times on page load)

sahilsaini432: It's called two times right now not 3 times, I will take a quick look at it.

fredy20501: I put a log message in the `getVersion` function in `api.js` and it shows up 3 times in the console, so for me it is called 3 times.

application/src/api/api.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable quotes */
+/* eslint-disable no-unused-vars */
```
fredy20501, line 2: In general, we shouldn't disable eslint rules for individual files. If you really think that a rule should be disabled you should do it globally in the eslint config file `.eslintrc.json`.

For this case, these 2 rules don't seem to be causing issues anyway so we should just remove these lines.

application/src/App.test.js
```diff
@@ -51,18 +51,4 @@ describe('Welcome Page', () => {
...
-    // Click the button
-    fireEvent.click(welcomeButton);
-    // After the button is pressed it should not be on the screen
-    expect(welcomeButton).not.toBeInTheDocument();
-
-    // Get homepage components from the screen
-    const homepage = screen.getByTestId('home-page');
-    // Check to see if the "Homepage" is on the screen
-    expect(homepage).toBeInTheDocument();
-  });
```
fredy20501, line 67: Why did you remove this test? If a test stops working we should fix it, not remove it.

sahilsaini432: This test requires the backend to be running at the same time. This will be covered in the integration test.

fredy20501: Oh because clicking the button does the auth token API call, right. Cool, that makes sense 👍

application/src/components/HomePage.js
sahilsaini432: Ya, it should not run 3 times. It should run only 2 times.
first time when all the components are rendered and the `getVersion` has no output. And the second time when you enter the Homepage from the Welcome Page after generating a valid auth token.

fredy20501: When you say "all the components are rendered", I don't think React renders all the components when the site first loads since you have the If/Then tags in `App.js` which means the Homepage component is only rendered once the Else block is triggered (i.e., the page property is changed).

I can actually see this since after putting a log line in `getVersion` I can see that it isn't triggered when the welcome page loads, but only it transfers to the home page after the button is pressed.

sahilsaini432: So, I updated it to call `useEffect` only one time, but the `getVersion` is still being called two times for some reason. When I look at the network call, it looks like one `getVersion` call is a `PREFLIGHT`, not really sure what that means.

fredy20501: For me it is only called once (after adding a logger.info line in the `getVersion` function I only see 1 console log). The preflight thing might be like some optimisation made by React in the background or something.

fredy20501: I don't think the preflight request is because it is called twice, so I don't think we need to worry about it:
https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

application/src/api/api.js
```diff
@@ -0,0 +1,48 @@
...
+  } catch(error) {
+    logger.warn(error);
+
+    if(error?.response?.status === 401) {
+      // Get New Auth Token
+      await getAuthToken();
+      return null;
+    }
+  }
+}
```
fredy20501, line 48: When we get a response status 401 we need to get a new auth token but more importantly, **we should try the request again** now that we have a good token (the frontend is going to need to do the request again so we might as well do it automatically).

I thought of 2 ways we could do this:
1. Put a for-loop around the content of the function and remove the `return null` so that it tries up to X times if it can't get the auth token.
2. Instead of `return null` do a recursive call to `getVersion` passing in a parameter representing how many times we have tries already, and adding a case to the function so that it stops once the number of tries reaches X.

(the number of retries X could be something like 2 or 3)

Note that we'll want to do this auto-retry for literally all API calls so we should abstract this into a function that can be used by all API calls.

Here's an implementation I did using a for-loop and having the abstracted function called `requestWithToken()`:
```js
// Get Versions
export const getVersion = async () => {
    return await requestWithToken(async() => {
        const response = await axios({
            method: 'GET',
            url: serviceUrl+'/version',
            headers: {
                token: authToken
            }
        });
        return response.data;
    });
}

const requestWithToken = async (request) => {
    for (let i=0; i<MAX_RETRY; i++) {
        try {
            // Make the given request
            return await request();
        } catch(error) {
            if(error?.response?.status === 401) {
                // Get New Auth Token and retry
                await getAuthToken();
            }
            else {
                // Don't retry if a different error occurs
                logger.warn(error);
                return null;
            }
        }
    }
    logger.warn('Unable to get auth token after '+MAX_RETRY+' tries.');
    return null;
}
```

I tested it and it seems to work fine, so we could use this if we want to.

sahilsaini432: I dont really understand why should we try request again automatically?
If the `authToken` is stale or invalid we generate a new one, now shouldnt the client make the request again instead of us automatically retrying the request again?

fredy20501: Yeah, the client will make the request again, but **we are the client** (this is the frontend code). This function doing it automatically means that instead of having to add a bunch of `if null then try again` all over our frontend code we only have this function that takes care of it.

sahilsaini432: so, instead of creating a separate function, we can also do something like this then, right?

````js
export const getVersion = async () => {
  try {
    retryCount = retryCount + 1;
    const response = await axios({
      method: 'GET',
      url: serviceUrl+'/version',
      headers: {
        token: authToken
      }
    });

    return response.data;
  } catch(error) {
    logger.warn(error);

    if(error?.response?.status === 401) {
      // Get New Auth Token
      await getAuthToken();
      if(retryCount === MAX_RETRY_LIMIT) {
        retryCount = 0;
        return null;
      } else {
        return await getVersion();
      }
    }
  }
}
````

Comment (fredy20501) [reply application/src/api/api.js:48]: Yep, exactly. That's the recursive way to do it.

The advantage with the separate function is that we don't have to re-write all of this code for each function that does an API call, we can just call the separate function and pass in the request we want to do. The way you just showed works perfectly fine but we'll have to copy this pattern for each function that does an API call, leading to a lot of repeated code, which isn't good.

Comment (sahilsaini432) [reply application/src/api/api.js]: Ya, that kind of makes sense. I will change to use the function.

**Review by fredy20501 — APPROVED** [review]
Nice work!

I tested manually to verify the functionality of the automatic retry for the API call.
- Works first try when token is valid.
- When token is invalid the call is done once, fails, then succeeds on the second try with the new token. (tested by restarting the frontend without going back to the welcome page)
- When the API cannot get a valid token it tries 3 times before returning null. (tested by modifying the back-end API to not save tokens)

File: application/src/api/api.js [hunk 9]
```diff
@@ -0,0 +1,63 @@
...
+
+// Auth Token
+let authToken;
+
+// Get environment
+require('dotenv').config();
+const ENV = process.env;
+
+// Retry Variables
+const MAX_RETRY_LIMIT = 3;
```
nathanielcaron, line 12: Should we only retry 2 times instead?
First try fails, then we get a token and try again.

I could see this creating unused auth tokens...

application/src/api/api.js
```diff
@@ -0,0 +1,63 @@
...
+          url: serviceUrl+'/version',
+          headers: {
+              token: authToken
+          }
+      });
+      return response.data;
+  });
+}
+
+const requestWithToken = async (request) => {
```
nathanielcaron, line 44: This is exactly what I had in mind.
The front end is aware that it needs a token and will retry automatically.
Plus the components themselves never have to worry about auth tokens as our API file handles it.

Nicely done.

nathanielcaron (approved):

application/src/api/api.js
fredy20501: 2 would work just fine, so I'm ok with that. If it fails twice that most likely means there's an underlying issue rather than just having an expired token, so trying a 3rd time isn't really necessary.

sahilsaini432 (commented):
Is this good to merge now??

fredy20501:
> Is this good to merge now??

Yep, you got 2 reviews & I tested it so we're good to merge.

PR #35, opened 2021-11-08
DBO-9 Added put endpoint for /userpost/:id, as well as get /userposts

JoaoPontes98:
How to update a userpost:
1.  Provide a .json object in the body with a single field called `update`. 
2.  Inside the `update` field, include whatever userpost properties you wish to update and the new data. Ex `{'update' : {'title':'new title'}}`
3. Send the put request to /userpost/:<doc_id>

How to get userposts:
1.  (Optional) include a filter in the body to return all userposts that match the filter. Ex `{filter: {authorID: <Auth_ID>}}`. If no filter is provided, all userposts will be returned
2.  Send the get request to /userposts

You will also notice some changes to the schema. Since this was to create a put endpoint, I made some of the fields immutable so that they can't be changed once the document has been created.


Service/api/api.raml
```diff
@@ -119,6 +136,21 @@ types:
           body:
             type: string
             example: User not found.
+        500:
+          body:
+            type: string
+    put:
```
nathanielcaron, line 142: Typically, if the `PUT` method is used, then the entire entity should be provided in the body of the request.
If only certain properties are provided, then the `PATCH` method should be used.

https://rapidapi.com/blog/put-vs-patch/

Service/api/api.js
```diff
@@ -108,7 +108,41 @@ exports.getUserPost = (req, res) => {
...
+  // empty filter returns all docs from Userposts
+  let filter = {};
+  if (req.body.filter) filter = req.body.filter;
+
+  UserPost.find(filter)
+    .then((docs) => {
+      // If the filter has no results
+      if (docs.length === 0 && filter !== {}) {
+        return res.status(406).send('No userpost matches the filter');
+      }
```
fredy20501, line 131: I'm not sure if we should consider no result as an error. We could just return an empty array/object instead (or maybe just return `docs` no matter what).

This decision will determine what the frontend needs to do. Either (a) check for status 406 and have a special case when empty or (b) check the length of the result and have a special case for that. 
Personally, I prefer (b) since I think we would still check the length of the result regardless, but it's up for debate.

fredy20501: Also, why do you check if the filter is not empty (`filter !== {}`)? (why is it only an error if the filter is not empty?)

JoaoPontes98: My thinking behind checking this was 'If there is a field in the filter that yields no results, then that means it's a bad filter and the developer should know.' But yeah looking back at it now, the logic is overly specific since just checking the length should be good enough for that.

I also agree that just sending docs would make it easier for programming the front-end. I'll be sure to remove this error.

Service/api/api.raml
JoaoPontes98: You guys cool with this being a PATCH instead then? I could quickly fix the raml/router.js to use PATCH.

Or is it best practice to have this act as PUT? That would involve tweaking the endpoint a bit but not much work at all.

fredy20501: I'm good with just changing it to PATCH. (given the planned functionality of our app I don't think we'll ever use this endpoint anyway)

JoaoPontes98:
I added the access_key generation in the CREATE. I also changed the delete endpoint to instead use the access key in the parameters instead of the document ID. Since we were never using the document ID to find the doc to delete, I thought this way made more sense.

fredy20501 (approved):
Nice work! 👍

Service/api/api.js
```diff
@@ -108,7 +95,35 @@ exports.getUserPost = (req, res) => {
...
+  let filter = {};
+  if (req.body.filter) filter = req.body.filter;
+
+  UserPost.find(filter)
+    .then((docs) => res.status(200).send(docs))
+    .catch((err) => {
+      logger.error(err);
+      return res.status(500).send(err);
+    });
+};
```
fredy20501, line 119: I tried testing these endpoints with Postman but ran into issues. 

First, GET endpoints don't have a body, so we can't access `req.body.filter` (always undefined). The (only?) way to pass data through a GET request is using URL encoded parameters (e.g., `https://google.com?param1=value1&param2=value2`) which are then accessed through `req.query` in the express endpoint. Now this is a potential problem since passing objects as URL parameters is not easy since it needs to be url-encoded by the frontend and then decoded back into a json object by the backend endpoint. I'm pretty sure it is possible, but it's not ideal.

Another (much simpler) solution would be to just change this endpoint to a POST so that we can pass the filter object in the request body. (this probably violates the REST API guidelines)

Note: the same issue happens with the PATCH endpoint since PATCH requests also don't have a body, apparently.

fredy20501 (changes requested):
Found issues during testing.

Service/api/api.js
JoaoPontes98: Damn thanks for bringing this up because I had no clue about this convention. After some research, it does seem that it is supported for GET requests to have body's but it's not really recommended to do so. ([link](https://stackoverflow.com/questions/978061/http-get-with-request-body)) Below, there's a screenshot of me running the get with a filter, and it does return the two posts with the same title as expected. 

So as far as I can tell, this current implementation does work. but it isn't the way we SHOULD be doing it. To be honest, when developing this, I saw that mongoose had this cool built-in `filter` feature that I thought was an easy way to implement queries with our get. But if we're going to do it right, making an endpoint to query like this should probably be its own ticket in the future.

I'd say to fix this, I could just remove the filter from the get userposts endpoint, and just leave it to return all userposts for now. We could tackle querying using `filter` on a later date.

As for PATCH's not having body's, I think you might be mistaken about that. As far as I can, I'm seeing sources showing that they do have body's. ([link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH))

image

sahilsaini432: Instead of using query can't we just the data as a `header` in the get request? like this
`const token = req.header('token');`

fredy20501: Oh, so I guess my problem was just me not understanding how to use postman properly work lol 😅 . 

For the GET body, I did see the same link you gave and I think we're fine keeping it as it is if it works. 
> A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.

Even though the standard says that a GET body doesn't have any defined meaning, since our API is not exposed externally (i.e., only our frontend will use it, it is not something the public will have access to) I think we're fine to deviate from the standard for simplicity.

fredy20501: The solution decided during the meeting was to change this endpoint to a POST and keep the filter object. (Nathaniel will do it)

nathanielcaron (approved):

Service/api/api.js
```diff
@@ -108,7 +108,41 @@ exports.getUserPost = (req, res) => {
...
+      }
+      return res.status(200).send(doc);
+    })
+    .catch((err) => {
+      logger.error(err);
+      return res.status(500).send(err);
+    });
+};
+
+exports.getUserPosts = (req, res) => {
```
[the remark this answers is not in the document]
nathanielcaron: I added a limit of 1,000 posts for now. We can look further into this in a separate ticket:
`DBO-38 Limit and paginate results from post /userposts endpoint`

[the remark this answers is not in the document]
fredy20501: Yep I agree. We'll add pagination later.

fredy20501 (approved):

PR #37, opened 2021-11-08
Simplify our CSS to mainly use bootstrap with small overrides

fredy20501:
**IMPORTANT**: this is merging into Nathaniel's `add-terms-conditions` PR, not `dev`.

I have created a SASS file to override the default bootstrap values (i.e., create our own theme). 
See the file `application/src/custom.scss` (currently I only override the primary colour since that's all we need for now).

This was done so that we have the colours we want for our app (the orange/yellow) but without losing all of the powerful stuff that bootstrap does for us (things like shading, colour changes on hover/click/focus, etc.).

After this change, I removed the now unnecessary CSS from App.css and component.css, leaving us with much cleaner CSS files.

**Note**: 
In any future CSS files if you want to use the primary colour of our app (#EC4038) you can just reference it using the CSS variable `var(--bs-primary)`. Example:
```css
.my-stuff {
    color:  var(--bs-primary);
}
```

React App - Google Chrome 2021-11-08 18-01-08

### DIFF
--- a/application/package-lock.json
+++ b/application/package-lock.json
@@ -10,6 +10,7 @@
         "@testing-library/jest-dom": "^5.14.1",
         "@testing-library/react": "^11.2.7",
         "bootstrap": "^5.1.1",
+        "node-sass": "^6.0.1",
         "react": "^17.0.2",
         "react-bootstrap": "^2.0.0-rc.0",
         "react-dom": "^17.0.2",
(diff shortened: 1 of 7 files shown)


sahilsaini432 (approved):
Nice 💯

nathanielcaron (approved):

nathanielcaron:
🥇

PR #38, opened 2021-11-09
Create map components

fredy20501:
I created 2 main map components and added them to the home page as a demo.

1. **LocationPickerMap**: Map that allows picking a location. 
**props**: only need to pass in a callback function that will be given the location when it changes (`onPositionChange`).

2. **StaticMap**: Static image of a map with a marker at the given position. Clicking on it opens a modal popup with a dynamic map with a marker at the same position.
**props**: width, height, position {lat, lng}, and zoom (optional).

**Functionality notes**:
- The location picker has a button to use the current location (this uses the standard geolocation API which most browsers support). If can't get the location, shows an appropriate message on the map.
- For the location picker you can set a location by clicking on the map or dragging the marker.
- The default location is New Brunswick for the location picker.
- The static map shows a placeholder image if it fails (e.g., if didn't give valid props to the component, making the URL invalid).

Images
---
Demo: (note: you can't see the browser popup asking for location permission in the screen capture)
React App - Google Chrome 2021-11-09 19-47-46

Location picker failing to get user's location:
image

Static image placeholder:
image


nathanielcaron:
Fantastic work! 🛩️ 
The `react-google-maps/api` package is really neat.

nathanielcaron (approved):

sahilsaini432 (approved):
Looks Great Fred 👍 💯

PR #39, opened 2021-11-11
DBO-5: Message Component

sahilsaini432:
Different Version of Message Component

Post Component
Screen Shot 2021-11-24 at 6 10 34 PM

Mobile View
Screen Shot 2021-11-24 at 6 11 28 PM


application/src/components/MainPage.js
```diff
@@ -0,0 +1,50 @@
...
+        Loading...
+      </div>
+    )
+  }
+
+  return (
+    <div id="home-page" data-testid="home-page">
+      <If condition={page === 'recent_posts'}>
+        <Then>
+          <Message />
```
fredy20501, line 43: I would rename the component to UserPost or Post to match the naming convention that we use in the database.

fredy20501 (changes requested):
I put some comments below for some specific sections, but there are also issues within the CSS that we'll need to fix (I just haven't looked too deep yet to tell you exactly what's missing since I don't have time right now). 

Here are some general things we'll need to look out for:
- Test what happens when each text field has a lot of text in it and add CSS to make it work as we want it to (also test for a very small amount of text, if needed - e.g. for the body).
- Test what happens with different sizes of things: lots of tags (does it wrap to another line?), very large/small image, very tall/wide image, etc.
- Make sure things are dynamic (limit the amount of hardcoded values, especially for width/height)
- I would consider redoing the main CSS from scratch rather than taking it from Figma since a) depending on the Figma page you used the layout might not have been "correct" (i.e., for some places I took shortcuts rather than doing it properly which means the generated CSS might be garbage, potentially) and b) I think a decent amount of our problems will come from this auto-generated CSS and it would be better to redo it using bootstrap-specific techniques & patterns. I can help you do this since I know the structure of the wireframes so I can redo it in CSS fairly easily.
- I'm not sure if the way the map and image move for smaller screens is the best. I would consider other ways of arranging them (maybe side by side or stretching the width  - I'd have to think about it and try some stuff).

application/src/components/post.css
```diff
@@ -0,0 +1,263 @@
...
+    border-radius: 20px;
+}
+
+#profile {
+    width: 100px;
+}
+
+#avatar {
+    width: 90px;
+    max-height: 90px;
```
fredy20501, line 24: The avatar is always a perfect circle, so the height should always be 90px (i.e., we should use `height` instead of `max-height`).

application/src/components/post.css
```diff
@@ -0,0 +1,263 @@
...
+
+#user-name {
+    height: 38px;
+    padding-top: 10px;
+    font-size: 18px;
+    line-height: 20px;
+    overflow-wrap: break-word;
+    hyphens: auto;
+    color: #000000;
+}
```
fredy20501, line 37: Can remove height and color here (they don't do anything).

application/src/components/post.css
```diff
@@ -0,0 +1,263 @@
...
+    line-height: 5px;
+    overflow-wrap: break-word;
+    hyphens: auto;
+    text-align: left;
+}
+
+#post-date {
+    position: relative;
+    bottom: 21px;
+    padding-right: auto;
```
fredy20501, line 70: We can remove `padding-right: auto;` (it isn't a valid value so it doesn't do anything).

application/src/components/post.css
```diff
@@ -0,0 +1,263 @@
...
+}
+
+#tag-container-sm {
+    display: none;
+}
+
+#tags {
+    padding: 2px 5px;
+    margin-right: 10px;
+    border: 1px solid #EC4038;
```
fredy20501, line 100: Replace the colour hex value with `var(--bs-primary)` (that's the primary colour of our app).

application/src/components/post.css
```diff
@@ -0,0 +1,263 @@
...
+        width: 40%;
+    }
+
+    #post-description {
+        width: 100%;
+    }
+
+    #avatar {
+        width: 100%;
+        height: 8%;
```
fredy20501, line 242: The image shouldn't have % dimensions. Instead should be fixed to a number of pixels. As it is currently the avatar image will stretch the longer the post is.

Example:
image

application/src/components/post.css
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    #post-date {
+        text-align: left;
+        padding-top: 10%;
+    }
+
+    #post-body {
+        padding-top: 9%;
+    }
```
fredy20501, line 186: Percentage padding makes things inconsistent and unprecise. Use a static padding instead (e.g., 20px for post body).

Note: there is another % padding further down that should also be fixed.

application/src/components/post.css
```diff
@@ -0,0 +1,263 @@
...
+        height: 100%;
+        border-radius: 1.5%;
+    }
+}
+
+/* For screens that are 992px or less */
+@media only screen and (max-width: 992px) {
+    #profile {
+        max-width: 18%;
+    }
```
fredy20501, line 157: Logically, the width of the profile column shouldn't be a percentage of the post width (i.e., if the post is wider, we wouldn't expect the profile column to take up more space), it should be a constant size. We should replace this with a static pixel measure (but still a different value for each `@media` query).

fredy20501 (changes requested):
I took a closer look at the CSS so I gave a lot of small change suggestions. I also got a few errors/warnings in my browser.

Sorry for all the comments 😅 you're probably getting tired of me at this point lol

fredy20501:
I would suggest putting the `post.js` and `post.css` files in their own folder to keep the components organized since it has its own CSS file (similar to what Nathaniel did with the privacy policy and terms&conditions).

fredy20501 (approved):

PR #40, opened 2021-11-21
DBO-22 - Create Karate API tests for the POST, GET, PATCH, DELETE endpoints for userPosts

nathanielcaron:
- Created Karate API tests
- Created .json data to use for karate API tests
- Updated API endpoints with better error handling


Service/api/api.js
```diff
@@ -77,13 +87,21 @@ exports.createUserPost = (req, res) => {
...
       }
-      return res.status(200).end();
+
+      UserPost.findByIdAndDelete(req.params.id)
+        .catch((err) => {
+          logger.error(err);
+          return res.status(500).send(err);
+        });
+      return res.status(200).send();
     })
```
fredy20501, line 105: The first query is `findOneAndDelete()` so the post is already deleted once you are in the `.then()` block. If you want to use a second independent query to delete (like you have added here) you should change the first one to just a `findOne()`.

nathanielcaron: This is a merging mistake from Joao's last PR

Service/api/api.js
```diff
@@ -38,14 +39,21 @@ exports.version = (req, res) => {
...
   const newUserPost = new UserPost({
     authorID: req.body.authorID,
     body: req.body.body,
     tags: req.body.tags,
     title: req.body.title,
-    imgURL: req.body.imgURL,
-    date_created: Date.now(),
+    img_URL: req.body.imgURL,
+    date_created: dateCreated,
     location: req.body.location,
```
fredy20501, line 57: The schema defines it as `imgURL` so if you want to change it to `img_URL` you should also change it in the schema.

Service/api/api.js
nathanielcaron: Updated.

Service/api/api.js
nathanielcaron: Updated all attributes to use underscores.

fredy20501 (commented):
We are missing tests for the `GET /userposts` endpoint (the `getUserPosts()` function, which takes in a filter object). Also, the error handling for that function should also be updated to match the others.

nathanielcaron:
> We are missing tests for the `GET /userposts` endpoint (the `getUserPosts()` function, which takes in a filter object). Also, the error handling for that function should also be updated to match the others.

That wasn't part of this ticket as we were thinking of having different ways of filtering posts.

I will create a separate ticket to take care of that.

fredy20501:
> > We are missing tests for the `GET /userposts` endpoint (the `getUserPosts()` function, which takes in a filter object). Also, the error handling for that function should also be updated to match the others.
> 
> That wasn't part of this ticket as we were thinking of having different ways of filtering posts.
> 
> I will create a separate ticket to take care of that. I can pick it up next.

Cool, that what I was thinking too. All is good 👍

fredy20501 (approved):
Nice work! 💯 ✔️

sahilsaini432 (approved):
NIce 💯

PR #42, opened 2021-11-24
Major CSS overhaul of Post component.

fredy20501:
**NOTE: THIS PR MERGES INTO SAHIL's PR (#39), NOT DEV**

Reworked the Post component so that the CSS is not conflicted between the old Figma CSS and bootstrap's CSS.

Major Changes:
- Created custom Row/Col elements instead of using bootstrap's ones to have more control over padding between rows/cols. (the custom ones I made are just simple divs with a flex-box style attached to them).
- Removed all unnecessary/redundant CSS.
- Improved flexibility of elements so they resize based on the overall size of the outer-container (which is controlled by bootstrap). This means that no element has a static size (except the avatar image).
- Changed CSS rules from ids to classes since multiple posts will be on a page at the same time (ids are for unique elements).
- Removed the global `font-size:20px` that was set on the App to instead rely on bootstrap's default font-size (which is in turn based on the browser's default font-size). If we want to make something with a different font size we should use `rem` relative sizes or the heading classes `h1` (biggest) through `h6` (normal).

### DIFF
--- a/application/src/App.css
+++ b/application/src/App.css
@@ -7,7 +7,6 @@
   flex-direction: column;
   align-items: center;
   justify-content: center;
-  font-size: 20px;
   color: black;
 }
 
--- a/application/src/components/FlexContainers.js
+++ b/application/src/components/FlexContainers.js
@@ -1,0 +1,38 @@
+// React
+import PropTypes from 'prop-types'
+
+// Stylesheet
+import './components.css';
+
+// These FRow and FCol components are just container divs with a specific class attached to them.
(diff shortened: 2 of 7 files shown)


fredy20501:
@sahilsaini432 Feel free to merge this (it is set to merge into your PR) if you are happy with it. Otherwise, just let me know if you have comments/suggestions/questions about what I did here.

sahilsaini432 (approved):
Nice 💯

PR #43, opened 2021-11-28
DBO-40 Karate API tests for the POST /userposts endpoint

sahilsaini432:


Automation/src/test/java/api/userPosts.feature
```diff
@@ -199,3 +199,131 @@ Feature: User post endpoints tests
...
+    And match response contains 'No Authentication Token Provided'
+
+  Scenario: Try to get a post with an invalid auth token
+    Given path 'userposts'
+    And header token = 'Invalid_Token'
+    When method post
+    Then status 401
+    And match response contains 'Invalid Authentication Token Provided'
+
+  # Test Cases for userposts endpoint
```
nathanielcaron, line 216: This comment should say POST userposts endpoint and be moved above the two tests added above

Automation/src/test/java/api/userPosts.feature
```diff
@@ -199,3 +199,131 @@ Feature: User post endpoints tests
     And request {access_key: post_access_key}
     When method delete
     Then status 200
+
+  Scenario: Try to get a post with no auth token provided
```
nathanielcaron, line 203: `Scenario: Try to get posts with no auth token provided`

Automation/src/test/java/api/userPosts.feature
```diff
@@ -199,3 +199,131 @@ Feature: User post endpoints tests
...
+    Given path 'userposts'
+    When method post
+    Then status 401
+    And match response contains 'No Authentication Token Provided'
+
+  Scenario: Try to get a post with an invalid auth token
```
nathanielcaron, line 209: `Scenario: Try to get posts with an invalid auth token`

Automation/src/test/java/api/userPosts.feature
```diff
@@ -199,3 +199,131 @@ Feature: User post endpoints tests
...
+
+    # Call userPosts endpoint with title as filter
+
+    Given path 'userposts'
+    And header token = auth_token
+    And request { filter: { title: "Ferrari" } }
+    When method post
+    Then status 200
+    And match response == '#[1]'
+    And match response[0].title contains "Ferrari"
```
fredy20501, line 261: Checking the number of posts (response length) to be a specific amount will make this test unstable. This is because cleanup is not guaranteed: if the test fails, the posts will not be deleted so the next time it runs there will be multiple copies of each post, meaning that the response size will be larger.

I have thought of 2 ways to fix this:
1. Check that all posts in the response match the filter (in this case, that all posts have title that contains "Ferrari"). I don't know if this is doable with Karate.
2. Find a way to do cleanup (I looked into it a little and there is an `AfterFeature` block we could use to delete stuff, but we'd have to know the access_keys so that's probably not gonna work).

Service/api/api.js
```diff
@@ -129,7 +129,21 @@ exports.getUserPost = (req, res) => {
...
+      if (req.body.filter.tags.length > 0) {
+        filter = {
+          tags: { $all: req.body.filter.tags }
+        };
+      }
+    }
+    else {
+      filter = req.body.filter;
+    }
+  }
```
fredy20501, line 146: We might want to filter using tags *and* other fields in the future, so to make that work you could keep the full `req.body.filter` and just replace the `tags` element. I think something like this might work:
```js
  if (req.body.filter) {
    filter = req.body.filter;

    if (req.body.filter.tags) {
      delete filter.tags;
      // If the tags are not empty, add them to filter
      if (req.body.filter.tags.length > 0) {
        filter.tags = tags: { $all: req.body.filter.tags };
      }
    }
  }
```
(You should add a test case to test this too, something like matching both a tag and the title, for example).

Service/api/api.js
```diff
@@ -129,7 +129,21 @@ exports.getUserPost = (req, res) => {
...
 
   // Limit the returned results to 1,000 user posts
   UserPost.find(filter).limit(1000)
```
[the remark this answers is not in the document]
nathanielcaron: Yep, the dev branch currently has 100 as the limit.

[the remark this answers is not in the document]
fredy20501: Ah, right! I didn't realize that this branch wasn't up to date with dev. All good 😄

Automation/src/test/java/api/userPosts.feature
nathanielcaron: For the 1st idea:
You can iterate over an array of items and check that they match a certain property.
One way of doing this -> `* match each response2 contains { price: '#(expected[_$.currency])' }`

For the 2nd idea:
The `AfterFeature` statement is very useful in certain scenarios, however for this, its safe to assume that if the tests fail to delete something, an `AfterScenario` would likely also fail.
One way of cleaning up everything is by doing it just before executing the tests (This is what I've done in the past). Although its probably not worth the time investment.

I vote for iterating over the response and checking their properties.

nathanielcaron (approved):

Service/api/api.js
```diff
@@ -129,7 +129,21 @@ exports.getUserPost = (req, res) => {
 exports.getUserPosts = (req, res) => {
   // empty filter returns all docs from Userposts
   let filter = {};
-  if (req.body.filter) filter = req.body.filter;
+
+  if (req.body.filter) {
+    // Check if the filters have tags
+    if (req.body.filter.tags) {
+      // If the tags are not empty, add them to filter
+      if (req.body.filter.tags.length > 0) {
```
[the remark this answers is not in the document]
fredy20501: Could be actually condensed to this: (Using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)).
`if (req.body?.filter?.tags?.length > 0) {...}`

fredy20501 (approved):

PR #44, opened 2021-12-01
DBO-26 Create Navigation Bar Component

fredy20501:
Work in progress.

image


application/src/components/Home.js
```diff
@@ -0,0 +1,84 @@
...
+  position: {
+    lat: 45.963589,
+    lng: -66.643112
+  },
+  location: 'New Brunswick, Canada',
+  postImage: PostImage,
+}
+
+// Homepage component for the application
+function HomePage() {
```
nathanielcaron, line 32: Should rename the component to `Home` here and on line 84 where it is exported.

nathanielcaron:
NIT: I think it might be worth looking into whether we can round the corners of the navbar in order for it to be consistent with the posts (Unless it is too much work).

fredy20501:
> NIT: I think it might be worth looking into whether we can round the corners of the navbar in order for it to be consistent with the posts (Unless it is too much work).

I actually already fixed that a while ago, just didn't push yet. 😄 (it is still a work in progress, after all)

sahilsaini432:
For the welcome page, isint the logo supposed to be in the center. Maybe it's just for me, but right now it's not vertically in the center.

Screen Shot 2021-12-27 at 1 10 29 AM (3)

application/src/components/navbar/navBar.css
```diff
@@ -0,0 +1,58 @@
...
+#navbar .nav-link.active {
+    font-weight: bold;
+}
+#navbar .nav-link:hover {
+    background-color: lightgray;
+}
+
+/* Aling items right in hamburger menu */
+#navbar .navbar-nav {
+    align-items: flex-end;
```
sahilsaini432, line 23: It's more of a suggestion, but would the hamburger menu look better as centred rather than flex-end?

Screen Shot 2021-12-27 at 1 15 21 AM

fredy20501: Yeah, that was the default. I changed it because to me I thought it looked better right-aligned since the hamburger button itself is also right-aligned. I have to admit that I don't really like either.

We could also do a pop-up menu instead. Something like this:
https://react-bootstrap.github.io/components/navbar/#navbar-offcanvas

sahilsaini432: Ya, off-canvas look a bit better

fredy20501:
> For the welcome page, isn't the logo supposed to be in the center? Maybe it's just for me, but right now it's not vertically in the center.

@sahilsaini432 Right, so I think I changed this when I did the CSS for the post component. The reason is that in the beginning it was centering our App.js component which meant that all of our pages were centered vertically, which was weird. I'll put it back  so it does make it centered but just for the welcome page. Nice catch! 👍

sahilsaini432 (approved):
Nice work Fred 👍 
Looks great 💯

application/src/components/navbar/navBar.css
nathanielcaron: I think the centered menu looks decent especially considering that this would be on a small phone screen where everything needs to be larger icons.

nathanielcaron:
NIT: I think it would look better if we could remove this grey outline around the hamburger menu.
Screen Shot 2021-12-29 at 11 59 44 PM

nathanielcaron:
Great work! Front-end components are much cleaner now.

application/src/components/navbar/navBar.css
fredy20501: Changed to the offcanvas popup.

nathanielcaron (approved):

sahilsaini432 (approved):

nathanielcaron:
Re-tested everything and it all looks great. 👍 

The only thing I think I would remove is the `Menu` text in the pop-over menu, don't think it serves much of a purpose.
Screen Shot 2022-01-03 at 5 18 10 PM

fredy20501:
> The only thing I think I would remove is the `Menu` text in the pop-over menu, don't think it serves much of a purpose.

@nathanielcaron Yeah, I also think that the "Menu" does look a little weird. I did find alternatives but I thought that it looked best with the "Menu" label.

### **We can either remove the label and have the X button left or right-aligned:**
image 
image

### **Or remove the header entirely (also removing the X):**
image

### **Conclusion**
I like the no header one (last one above) but I think not having the X makes it equally weird. I also don't like just the X without a label  since then there is an awkward white space with the X. Nevertheless, I don't have a big preferences so if someone really wants one version over another I don't mind. 🤷‍♂️ 🤔

nathanielcaron:
> > The only thing I think I would remove is the `Menu` text in the pop-over menu, don't think it serves much of a purpose.
> 
> @nathanielcaron Yeah, I also think that the "Menu" does look a little weird. I did find alternatives but I thought that it looked best with the "Menu" label.
> 
> ### **We can either remove the label and have the X button left or right-aligned:**
> image image
> 
> ### **Or remove the header entirely (also removing the X):**
> image
> 
> ### **Conclusion**
> I like the no header one (last one above) but I think not having the X makes it equally weird. I also don't like just the X without a label since then there is an awkward white space with the X. Nevertheless, I don't have a big preferences so if someone really wants one version over another I don't mind. 🤷‍♂️ 🤔

I like the left-aligned X button with no menu text and no header equal. Either it fine with me.

PR #45, opened 2021-12-21
DBO-10, DBO-25, DBO-39 - User Endpoints, Error Handling, Karate API Tests

nathanielcaron:
`DBO-39 - Make sure service NEVER crashes`:
  - Added pm2 (process manager) package
  - Added info in readme
  - Added global error handler in `app.js`
  - Changed all client and server errors to be in JSON format
 
`DBO-10 - Create the POST, GET, and DELETE endpoints for users`:
  - Created POST, GET, DELETE endpoints (No need for PATCH)
  - Added new endpoints to `RAML` file
  - Added routes to new endpoints in `routes.js`
  
`DBO-25 - Create Karate API Tests for User endpoints`:
  - Added Karate API tests for all 3 new endpoints in `user.feature`
  - Added user sample data to use in tests in `user.json` and `user_missingName.json`
  
`IMPORTANT`:
  - The process for creating user posts has changed. When a user is created, a new access key is generated. When a user post is created, this access key must be passed in the request body instead of being generated. This way, the user and the post will have the same access key.


Service/api/api.raml
```diff
@@ -151,10 +200,84 @@ types:
...
+              }
+
+/user:
+  post:
+    description: "Creates a new user document and returns it."
+    responses:
+      201:
+        body:
+          application/json:
+            type: user
```
fredy20501, line 219: Small typo: the type should be `User` with a capital U.

Automation/src/test/java/api/user.feature
```diff
@@ -0,0 +1,144 @@
...
+    Then status 201
+    * def user_id = response.user._id
+    * def user_access_key = response.user.access_key
+
+    # Call delete user endpoint
+    Given path 'user/' + user_access_key
+    And header token = auth_token
+    When method delete
+    Then status 200
+    And match response.message == 'User Deleted Successfully'
```
fredy20501, line 85: The description of this scenario doesn't match what it's doing. We probably want to rewrite this scenario to use an invalid access key and fail. The success case is covered with the final scenario of `Create, get, and delete a user`.

Service/api/api.raml
```diff
@@ -151,10 +200,84 @@ types:
...
+                  "message": "User Not Found"
+                }
+        500:
+          body:
+            application/json:
+              example: |
+                {
+                  "message": "An Unknown Error Occurred"
+                }
+    delete:
```
sahilsaini432, line 257: Shouldn't there be a `/{ak}` before `delete:`? 
just like `/{id}`

Service/routes/routes.js
```diff
@@ -9,74 +10,99 @@ const upload = multer({
 
 const API = require('../api/api');
 
+// Wrapper function to use global error handler
+const USE = (fn) => (req, res, next) => {
+  Promise.resolve(fn(req, res, next)).catch(next);
+};
```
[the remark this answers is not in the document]
nathanielcaron: This is where the global error handler in `App.js` comes in.
This function makes it so that if there is an error, then it will be handled by the global error handler, hence the .catch(next). You don't need the brackets but I can add them.

Service/api/api.js
```diff
@@ -204,14 +207,92 @@ exports.getUserPosts = (req, res) => {
...
+    return res.status(400).send({ message: 'No Request Body Provided' });
+  }
+
+  const accessKey = uuidv4();
+  const newUser = new User({
+    name: req.body.name,
+    avatar_url: req.body.avatar_url,
+    email: req.body.email,
+    access_key: accessKey
+  });
```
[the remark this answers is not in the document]
nathanielcaron: The reason why I changed this is so we never end up with posts that reference a non-existing user.

We cannot limit the deletion of a post and not limit the deletion of a user. 
If we decide that we need to require an access key to delete the post, then we need to require it when deleting the user too.

If we want to have the possibility of users having multiple posts, then we'll need to first check if there are posts referencing the user before deleting it. We could use a counter for that i.e., `posts: 2`

Service/api/api.raml
nathanielcaron: Currently, it is under `/{id}`.

`{id}` and `{ak}` are just placeholders, we do the same thing with user posts on line 142.

For correctness, we could update this and separate them out to be under `{ak}` instead of `{id}`.

README.md
```diff
@@ -59,6 +59,13 @@ To run the Express service locally, use the following commands inside of the Ser
 `npm install`\
 `npm run dev`
 
+To run the service using pm2, use the following commands inside of the Service directory:\
+`npm install`\
+`pm2 start app.js`
+
+To kill the process run by pm2, use the following command inside of the Service directory:\
+`pm2 stop app.js`
+
```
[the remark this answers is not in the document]
nathanielcaron: Added to README

Service/app.js
```diff
@@ -43,13 +43,22 @@ if (ENV.NODE_ENV === 'dev') {
   logger.info('Service logger initialized');
 }
 
+// Global error handler
+APP.use(function(err, req, res, next) {
```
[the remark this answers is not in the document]
nathanielcaron: Updated

Service/api/api.js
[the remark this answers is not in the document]
nathanielcaron: FYI: The delete user endpoint does not require an access key anymore.

Service/api/api.raml
nathanielcaron: Updated

Automation/src/test/java/api/user.feature
nathanielcaron: Updated

Service/api/api.raml
nathanielcaron: The delete user post endpoint is now under /{ak}

The delete user endpoint is still under /{id} since it uses the id and not an access key.

Service/api/api.js
```diff
@@ -241,18 +240,22 @@ exports.createUser = (req, res) => {
...
 };
 
 exports.deleteUser = (req, res) => {
-  const acessKey = req.params.ak;
+  const userId = req.params.id;
+
+  if (!ObjectId.isValid(userId)) {
+    logger.info('Invalid User ID');
+    return res.status(400).send({ message: 'Invalid User ID' });
+  }
```
fredy20501, line 255: We should add this case to the RAML file. (Invalid user ID, status 400)

fredy20501 (approved):
Nice work! 🥇

Service/api/api.js
nathanielcaron: Good catch. Updated.

sahilsaini432 (approved):
Nice Work 💯

PR #46, opened 2021-12-24
DBO-47 and DBO-38: Updated userposts endpoint to support partial tag match and recent posts.

sahilsaini432:


Service/api/api.js
```diff
@@ -178,34 +180,104 @@ exports.getUserPosts = (req, res) => {
...
+          location: 1,
+          maxTagMatch: {
+            $size: {
+              $setIntersection: ['$tags', providedTags]
+            }
+          }
+        }
+      },
+      { $sort: { maxTagMatch: -1 } }
+    ];
```
fredy20501, line 253: We also want to break tag-matching ties by sorting by most recent first since people will want to see more recent posts first. We should also add a final tie-breaker sort to prevent inconsistent orders using a unique field like `_id`.

This should work:
```js
 { $sort: { maxTagMatch: -1, date_created: -1, _id: 1 } }
```

fredy20501 (commented):
Nice work! I tested it locally and it worked like a charm. 💯

Service/api/api.js
```diff
@@ -178,34 +180,104 @@ exports.getUserPosts = (req, res) => {
...
-    });
+  let postData = [];
+
+  try {
+    if (recentPosts) {
+      postData = await UserPost.aggregate(searchFilters).limit(25);
+    }
+    else {
+      postData = await UserPost.aggregate(searchFilters).skip(pageNumber * 25).limit(25);
+    }
```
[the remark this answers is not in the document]
sahilsaini432: The first call is to get recent posts. The second one is for getting searched results. In the case of recent posts, we don't need to provide the page number because we not using skip at all. It will just compare the posts with the provided date to get the next page values.

This was something I wanted to ask in the next meeting, since this endpoint is getting way too complicated to you guys think it better to create a separate endpoint for the recent posts?

Service/api/api.js
```diff
@@ -178,34 +180,104 @@ exports.getUserPosts = (req, res) => {
...
+      { $sort: { date_created: -1 } }
+    ];
+
+    // Add the starting post date if provided (This is for paginating recent posts)
+    if (req.body.date) {
+      searchFilters = [
+        { $match: { date_created: { $lt: new Date(req.body.date) } } },
+        ...searchFilters
+      ];
     }
```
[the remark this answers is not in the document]
sahilsaini432: Pagination for searched results is handled by the `.skip()` function.  Since these are searched results sorting them by id or date would be weird as we want to display the post with the most matched tags first.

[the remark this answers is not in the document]
fredy20501: Oh yeah, you are completely right. I don't know what I was thinking 😅

Service/api/api.js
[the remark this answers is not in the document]
fredy20501: Yeah I mean my suggestion was to try to simplify it a bit but whether it is more "simple" is arguable, so it's fine to leave it as it is.

We should at the very least refactor the different parts into smaller functions and call those to make the overall endpoint shorter. We could also make separate endpoints too (but we need to think about what the frontend will need). We can definitely discuss it next meeting 👍

Service/api/api.js
```diff
@@ -178,34 +180,104 @@ exports.getUserPosts = (req, res) => {
...
+          title: 1,
+          body: 1,
+          tags: 1,
+          img_URL: 1,
+          date_created: 1,
+          location: 1
+        }
+      },
+      { $sort: { date_created: -1 } }
+    ];
```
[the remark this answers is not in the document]
sahilsaini432: Nice catch 👍

Service/api/api.raml
```diff
@@ -141,10 +141,11 @@ types:
 
 /userPosts:
   post:
-    description: Returns an array of Userposts which match a filter field passed in the req.body. An empty filter returns all userposts.
+    description: Returns an array of Userposts which match a filter field passed in the req.body. An empty filter returns all userposts with recent posts first (sorted by date).
     body:
       application/json:
-        example: { "filter": { "title" : "exampleTitle", "body" : "exampleBody"}}
+        example: { "filter": { "title" : "exampleTitle", "tags" : "["Tag1", "Tag2"]" } }
+        example: { "date": "date for the last displayed post" }
```
fredy20501, line 148: In the example for the date, we should put an actual date instead of a string description. Something like "2021-12-25".

nathanielcaron:
This is for `DBO-47 Update get /userposts endpoint to filter by partial tag match`
And `DBO-38 Limit and paginate results from post /userposts endpoint`.

Not `DBO-48`.

Service/api/api.js
```diff
@@ -178,34 +180,91 @@ exports.getUserPosts = (req, res) => {
...
+        ...searchFilters
+      ];
+    }
+  }
+  // Create new search format for partial matching tags
+  if (providedTags.length > 0) {
+    searchFilters = [
+      ...searchFilters,
+      {
+        $project: {
```
nathanielcaron, line 227: NIT: Why is this called project?

nathanielcaron:
Great work on this, this is exactly what I had in mind.

nathanielcaron (approved):

Service/api/api.js
sahilsaini432: `$project` is just a keyword used by MongoDB for creating an aggregation pipeline. This idea is that inside project, you can specify which value (data) to pick from a document. It is also needed to do any calculation. Like in this case we are using it to calculate the maximum number of tag matches and store the value in `maxTagMatch` so that we can use it to sort the documents in the next step.

This example might explain it better than me 😅 
https://docs.mongodb.com/manual/reference/operator/projection/positional/

Service/api/api.js
```diff
@@ -167,44 +170,85 @@ exports.getUserPost = (req, res) => {
...
-      searchFilters = { ...searchFilters, ...filter };
+      // Check id filters have Title
+      if (req.body.filter.title) {
+        searchFilters = [
+          ...searchFilters,
+          { $match: { title: req.body.filter.title } }
+        ];
+      }
     }
   }
```
fredy20501, line 194: We can remove the "filter" group/object from the body since it isn't really needed. (i.e., replace all `req.body.filter` by `req.body`). The frontend will provide the data directly in the body (currently we only use `tags`, `title`, and `page`).

Here's an example body to help clarify what I mean:
(before change)
```js
{ filter: { tags: ['test'], title: 'This is the title' }, page: 1 }
```
(after change)
```js
{ tags: ['test'], title: 'This is the title', page: 1 }
```

It should allow us to simplify the beginning of the function to something like this:
```js
exports.getUserPosts = (req, res) => {
  // empty filter returns all docs from Userposts
  let searchFilters = [];
  let providedTags = [];

  if (req.body.tags?.length > 0) {
    searchFilters = [
      { $match: { tags: { $in: req.body.tags } } }
    ];
    providedTags = req.body.tags;
  }

  // Check id filters have Title
  if (req.body.title) {
    searchFilters = [
      ...searchFilters,
      { $match: { title: req.body.title } }
    ];
  }

  // Create new search format for partial matching tags
  ...
```

Service/api/api.js
```diff
@@ -167,44 +170,85 @@ exports.getUserPost = (req, res) => {
 
 exports.getUserPosts = (req, res) => {
   // empty filter returns all docs from Userposts
-  let searchFilters = {};
+  let searchFilters = [];
+  let providedTags = [];
```
fredy20501, line 174: Since this endpoint requires a body, we can use the `if` block that is already used in other endpoints to check for a body:
```js
  // No request body provided
  if (!req.body || !Object.keys(req.body).length) {
    logger.info('No Request Body Provided');
    return res.status(400).send('No Request Body Provided');
  }
```
(This block is used in `updateUserPost()` and `createUserPost()`, so you can just copy-paste it)
*bonus points if you want to refactor that block of code into a function to reduce code duplication 🥇*

P.S. If you add this make sure to change the status code for the `'Invalid search filters provided'` error since it is also using status code `400`.

Service/api/api.js
```diff
@@ -167,44 +170,85 @@ exports.getUserPost = (req, res) => {
...
+  ];
+
+  if (req.body.date) {
+    searchFilters = [
+      { $match: { date_created: { $lt: new Date(req.body.date) } } },
+      ...searchFilters
+    ];
+  }
+
+  UserPost.aggregate(searchFilters).limit(POST_LIMIT)
```
fredy20501, line 247: Recent posts will also need to paginate, so we can copy the `req.body.page` check with the `.skip()` from the other function to here as well. (also update the `.raml` example)

Service/api/api.raml
```diff
@@ -141,10 +141,30 @@ types:
 
 /userPosts:
   post:
-    description: Returns an array of Userposts which match a filter field passed in the req.body. An empty filter returns all userposts.
+    description: Returns an array of Userposts which matches the filters provided in req.body.
     body:
       application/json:
-        example: { "filter": { "title" : "exampleTitle", "body" : "exampleBody"}}
+        example: { "filter": { "title" : "Post Title", "tags" : "["Tag1", "Tag2"]" }, "page": "1" }
```
fredy20501, line 147: Nit: (Small typo in the example) The array of tags shouldn't be in quotes.
```suggestion
        example: { "filter": { "title" : "Post Title", "tags" : ["Tag1", "Tag2"] }, "page": "1" }
```

Service/routes/routes.js
```diff
@@ -52,6 +52,12 @@ ROUTER.post(
   API.getUserPosts
 );
 
+ROUTER.post(
+  '/recentposts',
+  API.verifyAuthToken,
+  API.getRecentPosts
+);
```
fredy20501, line 59: The `/recentposts` endpoint doesn't have any Karate UI tests. You can either create tests within this PR or make a ticket of the Trello board to do it later.

Service/api/api.js
nathanielcaron: Since we need to return a status code and a message and it is only 3 lines of code I wouldn't make it into its own function.
But yes, this should definitely be included. 👍

Service/api/api.js
sahilsaini432: It's paginated right now, we don't need `skip` here because we are using `{ $match: { date_created: { $lt: new Date(req.body.date) } } }` this match call.
So, let's say if we are on page 3, we will provide the date of the last displayed post on page two in the request body. Then the next call for page three will return all the documents less than the provided date. using `skip` here might cause some delay depending on how many posts are there in the database.

Service/routes/routes.js
nathanielcaron: I can create a ticket for this.

Up to you @sahilsaini432 if you want to pick it up.

Service/api/api.js
fredy20501: *facepalm*
You are totally and completely right, I forgot! 😅

sahilsaini432: No Worries 👍

Service/api/api.js
sahilsaini432: Just to make sure, what would be the best error code for invalid search filters? 🤔

sahilsaini432: I think 406 would be the best one. Is that one good??

nathanielcaron: Error code: `400`

Message: `Invalid Request Body Format` or `Invalid search filters provided`

fredy20501: You know what, you can probably use the same error code `400` since the frontend won't really need to know the difference. (the message is there to differentiate anyway).

nathanielcaron: When it comes to error codes, keeping it general is good.
400 is a user error, 500 is a server error.
You can be more descriptive in error messages.

The most used status codes are 200, 201, 400, 403, 404, 500
Other status codes are rarely used.

fredy20501 (approved):

PR #47, opened 2022-01-04
DBO-50: Added Karate Tests for /recentposts endpoint

sahilsaini432:


Automation/src/test/java/api/recentPosts.feature
```diff
@@ -0,0 +1,101 @@
...
+    Then status 201
+    * def post4_access_key = response.post.access_key
+
+    # Calling recentPosts endpoint (with two more posts added)
+
+    Given path 'recentposts'
+    And header token = auth_token
+    When method post
+    Then status 200
+    Then assert response.length >= totalPosts
```
fredy20501, line 75: The check should be that the response should be *at least* 2 more. If it is directly equal, then it is not correct (since that means the new posts weren't returned).
```suggestion
    Then assert response.length >= totalPosts + 2
```

Automation/src/test/java/api/recentPosts.feature
```diff
@@ -0,0 +1,101 @@
+Feature: User post endpoints tests
+
```
fredy20501, line 2: We should add a separate scenario that tests the "less-than" date filter. For example, we can create 2 posts and make a request passing a specific date chosen so that only 1 post is returned and verify that *only* that 1 post was correctly returned.

Automation/src/test/java/api/recentPosts.feature
fredy20501: Actually, this won't work if the number of posts is already more than the limit (currently, 15).

sahilsaini432: That's why I added less than equal too if it's at its limit it will have to be equal to that. I guess we can leave it at this right now once we update the endpoints to carry page limit we can change it then.

fredy20501: A different way to test that the most recent posts are returned is to check whether the last 2 posts *specifically* were returned. (since the last 2 posts were created more recently, they should be returned).

So the scenario would go like this: (in addition to what you already have currently)
1. Create the first 2 posts (store their `_id`s)
2. Call `/recentposts` and make sure the 2 posts you just created were returned (by checking the `_id`s).
3. Create the second 2 posts (store their `_id`s)
2. Again, call `/recentposts` and make sure the 2 posts you just created were returned (by checking the `_id`s).

Automation/src/test/java/api/recentPosts.feature
sahilsaini432: For this one, I don't think checking less than the date would be a good idea. Since all the tests run in parallel we cant be sure that it will return only one post. I tried creating a similar test to check the order of the post and it didn't work.

fredy20501: Oh, right. So instead of checking exactly one post we could verify that all post returned have the date less than the one you specified. The scenario could go like this:
1. Create a post
2. Store the current date/time in a variable
3. Create a second post
4. Call the endpoint with the date you stored
5. Check that all posts returned have date <= variable
6. Check that the second post is not part of the result

Automation/src/test/java/api/recentPosts.feature
```diff
@@ -98,4 +110,56 @@ Feature: User post endpoints tests
...
+    And request { date: '#(post2_date)' }
+    When method post
+    Then status 200
+    * def before = $response[*].date_created
+    * copy after = before
+    * Collections.sort(after, Collections.reverseOrder())
+    And match before == after
+    * def result = true
+    * def checkOrder = function(post) { if(post.date_created > post2_date) { result = false } }
+    * karate.forEach(response, checkOrder)
```
fredy20501, line 148: Nit: The date comparison should check if it is greater **or equal** since an equal date would be a failure (since the mongoose filter is doing a strictly less than check). Like this:
`if(post.date_created >= post2_date`

fredy20501 (approved):
Nice work! 💯

nathanielcaron (approved):

PR #36, opened 2021-11-08
DBO-24 - Add modal on the welcome page to display terms and conditions and privacy policy

nathanielcaron:
140825366-10d11c9d-2fc1-47fc-8bd5-837e1e026e31

PR #48, opened 2022-01-04
Fix GitHub Actions slowdown

fredy20501:
I noticed that our GitHub actions were taking a lot longer (3-4min) than at the start of our project (<1min). I looked at the history of our actions and found that the jump from 1min to 4min was caused when **I** *unnecessarily* added the `node-sass` package in PR #36 🤦‍♂️ 😅 

Summary of work:
- Removed the unnecessary `node-sass` package from `application/package.json`
- Cleaned up dependencies in `application/package.json` and `Service/package.json`
- Added dependency caching for npm into our `main.yml` file

Now the pipeline again takes <1min 🎉


sahilsaini432 (approved):
Nice Catch Fred 💯

nathanielcaron (approved):

nathanielcaron:
🎖️

PR #49, opened 2022-01-04
DBO-52: Updated the Schema and all endpoints to return author along with the post

sahilsaini432:
…e post

### DIFF
--- a/Service/api/api.js
+++ b/Service/api/api.js
@@ -137,8 +137,6 @@ exports.deleteUserPost = (req, res) => {
 
   // Find the userpost with the matching access key
   UserPost.findOneAndDelete({ access_key: acessKey })
-    .populate('author')
-    .exec()
     .then((doc) => {
       if (!doc) {
         logger.info('User Post Not Found');
@@ -204,7 +202,7 @@ exports.getUserPosts = (req, res) => {
   }
 
   // Create new search format for partial matching tags
-  if (providedTags.length > 0) {
+  if (providedTags.length > 1) {
     searchFilters = [
       ...searchFilters,
       {
@@ -226,6 +224,13 @@ exports.getUserPosts = (req, res) => {
       { $sort: { maxTagMatch: -1, date_created: -1, _id: 1 } }
     ];
   }
+  else if (providedTags.length === 1) {
+    // When only one tag is provided sort by date_created
+    searchFilters = [
+      ...searchFilters,
+      { $sort: { date_created: -1, _id: 1 } }
+    ];
+  }
 
   // If the searchFilters are empty an invalid (or no) filter was provided
   if (searchFilters.length === 0) {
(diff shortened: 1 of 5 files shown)


Service/api/api.js
```diff
@@ -198,14 +204,15 @@ exports.getUserPosts = (req, res) => {
   }
 
   // Create new search format for partial matching tags
-  if (providedTags.length > 1) {
+  if (providedTags.length > 0) {
```
fredy20501, line 207: We should keep this check at > 1 since in the case of 1 tag we don't need to run this block. If there is only 1, the match that was already done to match *any* tag will already take care of it.
(i.e. the `{ $match: { tags: { $in: req.body.tags } } }` we do above)

Service/api/api.js
```diff
@@ -135,6 +137,8 @@ exports.deleteUserPost = (req, res) => {
 
   // Find the userpost with the matching access key
   UserPost.findOneAndDelete({ access_key: acessKey })
+    .populate('author')
+    .exec()
```
fredy20501, line 141: We don't need to populate here since this endpoint doesn't return the post.

Service/api/api.js
```diff
@@ -230,7 +237,10 @@ exports.getUserPosts = (req, res) => {
   const pageNumber = req.body.page ? (req.body.page - 1) : 0;
 
   UserPost.aggregate(searchFilters).skip(pageNumber * POST_LIMIT).limit(POST_LIMIT)
-    .then((docs) => res.status(200).send(docs))
+    .then((docs) => {
+      UserPost.populate(docs, 'author')
+        .then((combinedDocs) => res.status(200).send(combinedDocs));
+    })
```
fredy20501, line 243: Doing it this way means that we do 2 separate queries each time (the aggregate + another query for populate). A more efficient way is to use [`$lookup`](https://docs.mongodb.com/manual/reference/operator/aggregation/lookup/) + [`$unwind`](https://docs.mongodb.com/manual/reference/operator/aggregation/unwind/#-unwind--aggregation-) within the aggregate pipeline so that everything is done within 1 query.

This [StackOverflow answer](https://stackoverflow.com/a/53404181) explains it very well, but this is basically what we need:
```js
// Create new search format for partial matching tags
  if (providedTags.length > 0) {
    searchFilters = [
      ...searchFilters,
      {
        $project: {
          title: 1,
          body: 1,
          tags: 1,
          author: 1,
          img_URL: 1,
          date_created: 1,
          location: 1,
          maxTagMatch: {
            $size: {
              $setIntersection: ['$tags', providedTags]
            }
          }
        }
      },

      // vvvvv THIS IS THE NEW STUFF vvvvv
      {
        $lookup: {
          from: User.collection.name,
          localField: 'author',
          foreignField: '_id',
          as: 'author'
        }
      },
      { $unwind: '$author' },
      // ^^^^^ THIS IS THE NEW STUFF ^^^^^

      { $sort: { maxTagMatch: -1, date_created: -1, _id: 1 } }
    ];
  }
```

P.S. Can do the same for the `getRecentPosts`.

Service/api/api.js
```diff
@@ -118,6 +118,8 @@ exports.updateUserPost = (req, res) => {
 
   const query = { _id: userPostId };
   UserPost.findOneAndUpdate(query, req.body.update, { new: true })
+    .populate('author')
+    .exec()
```
fredy20501, line 122: We should add a test for each endpoint that was modified here to check whether the user data is correctly populated.

nathanielcaron (approved):

fredy20501 (approved):
This works but there is a potential issue since we are now using the user IDs of specific users within the tests which are not created by the test. This means that if we clear the database the tests will fail.

We can keep this for now and fix it in a later PR once we figure out a solution. We can discuss this next meeting.

Possible solution: we could make sure that a specific user is created before any tests are run and deleted at the end. This could be done using Karate [`karate.callSingle()`](https://github.com/karatelabs/karate#karatecallsingle) to run a setup before all tests and use Junit's [`@AfterAll`](https://junit.org/junit5/docs/5.6.1/api/org.junit.jupiter.api/org/junit/jupiter/api/AfterAll.html) annotation to cleanup after at the end.

PR #51, opened 2022-01-07
DBO-27: Delete Post Component (+CSS cleanup)

fredy20501:
Summary of work
===
1. Cleaned up some CSS in the frontend.
    - I couldn't resist. 😅 
2. Modified the hover animation of the navbar and the post's tags & map.
3. Modified `get` & `delete` user endpoints to work with the frontend.
    - Now `get` uses the `access_key` and `delete` uses the `_id` (it was previously reversed).
4. Modified the user post database schema to work with the frontend.
    - Changed `latitude` and `longitude` to `lat` and `lng`.
    - Added `location_string`
5. Modified the Post component to work with the format given by our APIs.
6. Created a confirmation modal component (used to confirm deletion).

Tip: It might be easier to review each commit separately since I did changes in a lot of files.

New Hover Animations
===
Before             |  After
:---------------:|:---------------------:
  |  

Delete Component - Walkthrough
===

Edit: forgot to show the "Post not found" case:


nathanielcaron:
This looks fantastic!

The only thing I'm not sure about is the font size for the alert messages for when a post is deleted successfully and when it fails to delete (green and red alerts). It seems a little large to me.

nathanielcaron:
Is there something that was not working with the previous implementation of the delete user post endpoint where it used an access key?

The point of using an access key was so that we never expose an actual database object id to users of the application.

Automation/src/test/java/data/userPost.json
```diff
@@ -5,8 +5,9 @@
     "body": "Went hiking in a beautiful trail in Cape Breton with the family.",
     "img_URL": "https://catalogue.novascotia.com/ManagedMedia/27778.jpg",
     "location": {
-        "latitude": 0,
-        "longitude": 0
+        "lat": 0,
+        "lng": 0
     },
+    "location_string": "New Brunswick, Canada",
```
nathanielcaron, line 11: I assume the Google API uses this format in some way?

fredy20501: Nope, that is the string that shows up below the map on the post.

fredy20501:
> The only thing I'm not sure about is the font size for the alert messages for when a post is deleted successfully and when it fails to delete (green and red alerts). It seems a little large to me.

Yeah, it does look a bit big, especially on a small screen size (like I used for the demo). It's basically the same size as the titles. I'll try it with a "regular" font size and see how it looks.

application/src/api/api.js
```diff
@@ -41,6 +41,39 @@ export const getVersion = async () => {
...
+            'token': authToken
+        }
+    });
+    return response.data;
+  });
+}
+
+// Delete post
+export const deletePostByID = async (_id) => {
+  // TODO: Call the exposed delete post endpoint
```
nathanielcaron, line 60: `DBO-57 Call the exposed delete user post endpoint from the frontend when deleting a user post`

fredy20501:
> Is there something that was not working with the previous implementation of the delete user post endpoint where it used an access key?
> 
> The point of using an access key was so that we never expose an actual database object id to users of the application.

So my view is that deleting elements in the db should be done by `_id` since that is the only absolutely guaranteed unique field (there's even a specific deleteByID method because that is so commonly used). And yes, users don't see the `_id` so it isn't really "exposed".

> I assume this was done so the user could view the post before deleting it?

Well not technically. I changed the `get` request to use the access key because we want to display the post first, yes. But technically the delete post could have been kept the same (see above for my reason).

nathanielcaron:
> > Is there something that was not working with the previous implementation of the delete user post endpoint where it used an access key?
> > The point of using an access key was so that we never expose an actual database object id to users of the application.
> 
> So my view is that deleting elements in the db should be done by `_id` since that is the only absolutely guaranteed unique field (there's even a specific deleteByID method because that is so commonly used). And yes, users don't see the `_id` so it isn't really "exposed" (that's why we give users an access ID which they enter into the input field).
> 
> > I assume this was done so the user could view the post before deleting it?
> 
> Well not technically. I changed the `get` request to use the access key because we want to display the post first, yes. But technically the delete post could have been kept the same (see above for my reason).

That's a fair point. This allows us to have users find their post by access key and delete it by id. Good solution.

application/src/components/Delete.js
```diff
@@ -1,17 +1,166 @@
...
+            </Alert>
+          </Container>
+        </Case>
+
+        <Default></Default>
+      </Switch>
+
+      <ConfirmationModal
+        title="Confirmation"
+        body={
```
nathanielcaron, line 151: NIT: Could the body of the Confirmation modal instead already contain a div and only a string would need to be passed as a prop?

nathanielcaron:
Awesome work on this. 🎆  The confirmation modal and delete post components in particular are very clean.

application/src/components/Delete.js
fredy20501: This way is more flexible since it would allow you to put anything (e.g., multiple elements) and not just a simple string. The whole reason why I made it a separate component was to allow reuse, which is why the props are made to be as flexible as possible.

fredy20501: Edit: Changed it to use `children` prop since that makes more sense to me in hindsight. How's that? 😉

application/src/api/api.js
fredy20501: You got it, boss ✔️

sahilsaini432 (approved):
Nice work Fred 👍 
Works Perfectly 💯

JoaoPontes98:
Clean work Fred! Also good points on switching the ak and _id's

JoaoPontes98 (approved):

CarterMacLennan (approved):
Looks great to me, no complaints. Great work :-)

PR #52, opened 2022-01-07
DBO-32 - Create exposed "Create User Post" endpoints

nathanielcaron:
`DBO-32 Create exposed "Create User Post" endpoints`
- Added two new endpoints to `api.js`
- Added two new endpoints to `api.raml`
- Edited `routes.js` to hide dev endpoints when not in dev environment

`DBO-53 Create Karate API tests for the exposed create user posts endpoints`
- Divided userPost tests into four files (post, get, patch, delete)
- Added API tests in `userPost-post.feature` for two new endpoints


Service/api/api.js
```diff
@@ -452,3 +450,168 @@ exports.deleteImage = async (req, res) => {
       return res.status(500).send({ message: INTERNAL_SERVER_ERROR_MSG });
     });
 };
+
+exports.uploadPostImages = async (req, res) => {
+  // Request body must contain two images
+  if (!req.files || req.files.length < 2 || !req.files.avatar || !req.files.picture) {
```
sahilsaini432, line 456: If the user doesn't provide any image with the post (or no avatar), I am guessing we would have to assign a random image to the post then. Since this endpoint requires two images?

nathanielcaron: That is a good point.

There could be a user with no avatar and a post picture or vice versa.
I can update this endpoint to accept 1 or two images.

If there are no images provided however (no avatar and no post picture) then the frontend should not call this endpoint.

nathanielcaron: Updated the endpoint to accept both an avatar and post picture, only an avatar, or only a post picture.
The karate API tests have been updated as well to test all 3 scenarios.

Service/api/api.js
```diff
@@ -452,3 +449,195 @@ exports.deleteImage = async (req, res) => {
...
+  const {
+    user, post, avatarId, pictureId
+  } = req.body;
+
+  // Create user with uploaded avatar
+  const newUser = new User({
+    name: user.name,
+    avatar_url: BUCKET_URL + avatarId,
+    email: user.email
+  });
```
fredy20501, line 560: I believe we need an if-statement to check if `avatarId` is undefined since if it is the URL will be set to 
```js
BUCKET_URL+"undefined"
``` 
as a valid string. That's what happens when you append undefined to a string, apparently 😐

Service/api/api.js
```diff
@@ -452,3 +449,195 @@ exports.deleteImage = async (req, res) => {
...
+      deleteFile(avatarId)
+        .catch((avatarError) => {
+          logger.error(avatarError.message);
+        });
+
+      // Delete picture
+      deleteFile(pictureId)
+        .catch((pictureError) => {
+          logger.error(pictureError.message);
+        });
```
fredy20501, line 576: It doesn't feel right to delete the images in this endpoint when they were created in a different endpoint. From the frontend's perspective, this is not intuitive. I think that since we have separated the upload image step from the create post step, that the frontend should be responsible for calling the delete image endpoint in case of failure. (Ideally, those two steps would have been done together within one endpoint, in which case we wouldn't have this problem - btw, was there an issue preventing us from doing that? Can't we just cascade the middleware and have both functions called on the same endpoint?)

Example: If create post fails because of validation, I think it would be reasonable to expect the frontend to call create post again without having to re-upload the images.

Service/api/api.js
```diff
@@ -452,3 +449,195 @@ exports.deleteImage = async (req, res) => {
...
+          body: newUserPost.body,
+          tags: newUserPost.tags,
+          title: newUserPost.title,
+          img_url: newUserPost.img_url,
+          date_created: newUserPost.date_created,
+          location: newUserPost.location,
+          true_location: newUserPost.true_location,
+          access_key: newUserPost.access_key
+        }
+      });
```
fredy20501, line 640: Couldn't we just do this?
```js
return res.status(201).send(newUserPost);
```
(This is the same as the `createPost` function, so if you change it here we might as well change it there too)

Service/api/api.raml
```diff
@@ -26,7 +26,7 @@ types:
     properties:
       _id:
         type: number
-      author_ID:
+      author:
```
fredy20501, line 29: We also need to change `img_URL` to `img_url` below.

nathanielcaron: Updated

Service/api/api.js
nathanielcaron: Not in this case since we need to assign author: newUser

Service/api/api.js
nathanielcaron: Updated both avatar_url and img_url

Service/api/api.js
```diff
@@ -452,3 +449,195 @@ exports.deleteImage = async (req, res) => {
...
+    return res.status(400).send({ message: 'Missing User' });
+  }
+  if (!req.body.post) {
+    logger.info('Missing User Post');
+    return res.status(400).send({ message: 'Missing User Post' });
+  }
+  if (!req.body.avatarId || !req.body.pictureId) {
+    logger.info('Missing Image IDs');
+    return res.status(400).send({ message: 'Missing Image IDs' });
+  }
```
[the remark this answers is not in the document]
nathanielcaron: Removed

Service/api/api.js
```diff
@@ -452,3 +449,195 @@ exports.deleteImage = async (req, res) => {
...
+          return res.status(500).send({ message: INTERNAL_SERVER_ERROR_MSG });
+        }
+
+        const pictureKey = pictureResult.key;
+
+        const response = { pictureId: pictureKey };
+        return res.status(201).send(response);
+      });
+  }
+};
```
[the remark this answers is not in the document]
nathanielcaron: Updated

Service/api/api.js
```diff
@@ -452,3 +449,195 @@ exports.deleteImage = async (req, res) => {
...
+
+  // Create user with uploaded avatar
+  const newUser = new User({
+    name: user.name,
+    avatar_url: BUCKET_URL + avatarId,
+    email: user.email
+  });
+
+  newUser.save((userError) => {
+    if (userError) {
```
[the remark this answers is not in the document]
nathanielcaron: From my point of view, those other endpoints allowed us to gradually work our way up to this. Working on these endpoints without any prior code to go off of would have taken a lot more time.

It also allows us to test our service and database at a more granular level.

Re-using these functions would not work as-is since they are returning responses. I thought refactoring those functions was not worth the time.

Service/api/api.js
nathanielcaron: That is a good point. We can create a separate endpoint to delete both images together.

As for now calling multiple functions as middleware, this is the same issue I described in our meeting. We cannot pass both multipart form data and JSON data in a single request, unfortunately.

fredy20501: I don't think we need another endpoint, we could just let the frontend call the delete image one twice. (Or modify the delete endpoint so it takes an array of images, maybe?).

Right, I had forgotten what the specific issue was. That sucks :(

Service/api/api.js
fredy20501: I mean, the frontend doesn't even use/need the response so is that really necessary in the first place?

Service/api/api.js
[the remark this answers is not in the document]
fredy20501: Yep, that makes sense. I feel like there should be a better way to do this kind of stuff in general, though. Like, do you really think that professionals in companies and whatnot do this kind of process? I don't think so, but it doesn't feel worth it for us to re-think everything at this stage anyway, so I think we're good.

Service/api/api.js
nathanielcaron: Technically no, but since this endpoint is responsible for creating a user and a user post, it is good practice to return what was created. 

We can definitely update the other create post endpoint though.

fredy20501: Meh, it was just to save a few lines of code. Not really important.

Automation/src/test/java/api/userPosts-post.feature
```diff
@@ -362,15 +362,6 @@ Feature: User post endpoints tests
...
 
-  Scenario: Try to create a user and user post with missing picture id
-    # Call create user post endpoint with missing picture id
-    Given path 'post'
-    And header token = auth_token
-    And request read('../data/user_userPost_missingPictureId.json')
-    When method post
-    Then status 400
-    And match response.message == 'Missing Image IDs'
-
```
fredy20501: I would suggest keeping this to test that only providing one image (just avatar in this case) should still work. We can change the expected status code to 201 and delete the post & user after like the valid case (and maybe also check that the returned user object doesn't have an `avatar_url` property, in this case).

Service/api/api.js
nathanielcaron: Realistically, the frontend should try again to create the user and user post if uploading the images succeeded. If it still fails, then yeah, it should call the delete image endpoint for both images.

I'll create a small ticket to look into modifying that endpoint to accept an array of images to delete.

Automation/src/test/java/api/userPosts-post.feature
nathanielcaron: That's a good test case.

I added two tests for when either an avatar is not provided or a post picture is not provided.

Service/api/api.js
fredy20501: Cool, so are you going to remove these `deleteFile()` from this function?

nathanielcaron: Yes, will do. 👍

fredy20501 (approved):

sahilsaini432 (approved):
Works Great 👍 
Nice Job 💯

PR #50, opened 2022-01-07
DBO-33 API tests for image endpoints

JoaoPontes98:
The test coverage includes:
/image
- post no auth token
- post invalid auth token
- post with no file
- post success

/image/id
- get with invalid ID
- get no auth token
- get with an invalid auth token
- get success
- delete invalid ID
- delete success

/imageurl/id
- get with invalid ID
- get no auth token
- get with an invalid auth token
- get success

PR #54, opened 2022-01-10
DBO-48/49 : Exposed delete endpoint with its karate tests.

sahilsaini432:


Automation/src/test/java/api/deletepost.feature
```diff
@@ -0,0 +1,118 @@
+Feature: Delete post endpoint tests
```
nathanielcaron, line 1: I think these tests should instead be included in the `userPosts-delete.feature` file

Service/api/api.js
```diff
@@ -451,6 +451,100 @@ exports.deleteImage = async (req, res) => {
...
+      return res.status(500).send({ message: INTERNAL_SERVER_ERROR_MSG });
+    }
+    if (user === undefined) {
+      fullyDeleted = false;
+      status = {
+        user: 'User Not Found'
+      };
+    }
+
+    // Delete avarat img if provided
```
nathanielcaron, line 495: typo:
`// Delete avatar image if provided`

Service/utils/utils.js
```diff
@@ -35,3 +38,48 @@ exports.createImage = async (file) => (
       return null;
     })
 );
+
+exports.deleteImage = async (image) => (
+  // Delete image
+  deleteFile(image)
+    .then(() => {
+      logger.info('Image Deleted Successfully');
```
nathanielcaron, line 46: Don't think we need this log

Service/utils/utils.js
```diff
@@ -35,3 +38,48 @@ exports.createImage = async (file) => (
...
+      return null;
+    })
+);
+
+exports.deleteUser = async (userID) => (
+  User.findByIdAndDelete(userID)
+    .then((doc) => {
+      if (!doc) {
+        logger.info('User Not Found');
+        return undefined;
```
nathanielcaron, line 60: Not a big fan of this `undefined` vs `null` business...
Maybe return `{}` vs `null`?

I wonder if anyone has an alternate solution 🤔

sahilsaini432: We could just return a string saying `Not Found`?
That would look better.

nathanielcaron: @JoaoPontes98 @fredy20501 @CarterMacLennan

Automation/src/test/java/api/userPosts-delete.feature
```diff
@@ -53,3 +57,113 @@ Feature: Delete user post endpoint tests
...
+    And match response.status.postImg == 'Post Image Not Found'
+
+  Scenario: Try to fully delete the post
+    Given path 'deletepost/' + postID
+    And header token = auth_token
+    When method delete
+    Then status 201
+    And match response.status.message == 'Post Deleted Successfully'
+    And match response.post._id == postID
+    And match response.post.author._id == authorID
```
fredy20501, line 169: This is a problem. We are using this scenario as a "cleanup" to delete the post that was originally created by the `createPost.feature` but that creates a race condition since the scenarios are run in parallel, meaning that it is possible that the post gets deleted before the other scenarios that need it are done, causing intermittent errors.

Given that in this file we only use the post created by the `createPost.feature` in this scenario and the invalid token scenarios, I don't think it is useful to create a global post. For the auth token scenarios you don't have to pass in a valid `_id` which leaves just the final delete scenario, so you might as well just create it in that scenario specifically. (you might be able to do the `callonce read` just within this scenario, which would be great, but I don't know)

In general, the only time a global post is useful is when you use it **without modifying/deleting it**. In that case it can be reused since you know it won't change while the scenarios run in parallel.

Automation/src/test/java/api/userPosts-delete.feature
```diff
@@ -2,8 +2,12 @@ Feature: Delete user post endpoint tests
 
   Background:
     * url baseUrl
+    * def baseImageURl = 'https://senior-design-img-bucket.s3.amazonaws.com/'
     * def auth = callonce read('classpath:utils/authentication.feature')
+    * def post = callonce read('classpath:utils/createPost.feature')
     * def auth_token = auth.response.token
+    * def postID = post.response.post._id
+    * def authorID = post.response.post.author._id
```
fredy20501, line 10: We have to remember to clean up what was created by the `createPost.feature` at the end. We should be able to use something like this (in the `Background` block) to call a cleanup feature at the end: (but I don't know if you can pass the postID to it, so it might not work)
```cucumber
* configure afterFeature = function(){ karate.call('cleanup.feature'); }
```
Or maybe this:
```cucumber
* configure afterFeature =
"""
function(){
  // Do cleanup stuff here
}
"""
```

Service/api/api.raml
```diff
@@ -369,6 +369,51 @@ types:
...
+                }
+                post: user post
+              }
+        400:
+          body:
+            application/json:
+              example: |
+              {
+                message: "Invalid Post ID Provided"
+              }
```
fredy20501, line 401: The examples need to have the status object too, to match what the API is actually returning.
```raml
example: |
{
  status: { message: "..." }
}
```
(multiple places to fix this)

Service/api/api.js
fredy20501: Do a search and replace for `avarat` because there is also a variable with that typo a little bit lower down.

Service/api/api.raml
fredy20501: Nope I'm wrong

Service/api/api.js
```diff
@@ -451,6 +451,100 @@ exports.deleteImage = async (req, res) => {
...
+    status = {
+      user: 'No User Provided'
+    };
+  }
+  else {
+    // Delete user
+    const user = await UTILS.deleteUser(post.author._id);
+    if (!user) {
+      return res.status(500).send({ message: INTERNAL_SERVER_ERROR_MSG });
+    }
```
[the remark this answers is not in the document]
sahilsaini432: My bad, I thought that was only for the `404` error.

Service/api/api.js
```diff
@@ -451,6 +451,100 @@ exports.deleteImage = async (req, res) => {
...
+        };
+        fullyDeleted = false;
+      }
+    }
+  }
+
+  // Delete post image if provided
+  if (post.img_url) {
+    // Get post image id
+    postImageID = post.img_url?.substring(post.img_url?.lastIndexOf('/') + 1);
```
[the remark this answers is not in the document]
nathanielcaron: I agree. Something like `getImageId()` in utils could easily be unit tested.

Automation/src/test/java/api/userPosts-delete.feature
nathanielcaron: We can discuss this on Friday.

Service/utils/utils.js
fredy20501: A good way to solve this problem (in general, regardless of programming language) is to define an enum that has values representing the different possible outcomes and use those for returning from the functions and then again for comparing the function results. It also has the added benefit of increased readability since the enum values have descriptive names.

Javascript doesn't really have enums but we can define something like an enum like so:
```js
// Define enum
const ResultEnum = { Success: 1, NotFound: 2, Error: 3 };
Object.freeze(ResultEnum); // Freeze prevents modification of the object

// Export it
exports.ResultEnum = ResultEnum;

// In the Util functions:
return ResultEnum.NotFound;
return ResultEnum.Error;
return ResultEnum.Success; // (note: in our case we usually return a doc/object on success, so no need for those cases)

// When comparing the result of Util functions
const result = UTIL.createImage();
if (result === ResultEnum.Error) { ... }
else if (result === ResultEnum.NotFound) { ... }
```

Automation/src/test/java/api/userPosts-delete.feature
sahilsaini432: I don't think the race condition will be a problem here. The was its set up right now I think it runs feature files in parallel and the scenarios within them in sequence. I might be wrong about that 😅
@nathanielcaron Does it run feature files in parallel or Scenarios?

Automation/src/test/java/api/userPosts-delete.feature
nathanielcaron: I would lean towards calling a feature file instead of using a JS/Java function.

We can discuss the general approach on Friday, but I think we should do the following:

When we need a 'global post/user' or something that needs to be used for several tests, we should do a callonce in the background to create it and another call afterFeature to delete whatever was created.
It is ok to have multiple feature files create data to be reused. If anything, it just tests the robustness of our API more.

Automation/src/test/java/api/userPosts-delete.feature
fredy20501: I'm pretty sure it does run scenarios in parallel by default:
image

Service/utils/utils.js
CarterMacLennan: I typically just used {},  but I really like the idea of Fred's enum solution. Never saw that before.

nathanielcaron: That is a good solution that could be used everywhere.

If we have `eslint no-const-assign` turned on as error, then we don't need the `object.freeze()`.

fredy20501: Nathaniel `const` only prevents assignment. Object properties can still be modified since you aren't re-assigning the variable. It is a weird quirk of javascript.
Example
```js
const obj = {hello: 1, goodby: 2};
obj = somethingElse; // Invalid
obj.hello = 5; // Valid <- this is what freeze prevents (along with adding/removing extra properties)
```

Service/api/api.js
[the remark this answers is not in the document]
CarterMacLennan: Great idea, I second this

Automation/src/test/java/api/userPosts-delete.feature
nathanielcaron: https://github.com/karatelabs/karate#parallel-execution

`The parallel runner will always run Feature-s in parallel. Karate will also run Scenario-s in parallel by default.`

Automation/src/test/java/data/user_userPost_missingImages.json
```diff
@@ -0,0 +1,18 @@
+{
+  "post": {
+    "author": "1234",
```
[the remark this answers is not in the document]
sahilsaini432: Looks like I need to assign a temp value before I can use `set` to change it. If I remove `author` from here the `set` keyword doesn't seem to work. 🤔

JoaoPontes98:
I just wanted to document this - but with my other pr for DBO-33, we could refactor the image tests to just use one of the images included in this pr's `src/data` folder to reduce space. Not super important but a quick change nonetheless

sahilsaini432:
> I just wanted to document this - but with my other pr for DBO-33, we could refactor the image tests to just use one of the images included in this pr's `src/data` folder to reduce space. Not super important, but it's a quick change in the future nonetheless.

That is a good idea, I will change that in the next pr, once #50 and #54 are merged.

Automation/src/test/java/api/userPosts-delete.feature
```diff
@@ -53,3 +54,151 @@ Feature: Delete user post endpoint tests
...
+
+    Given path 'deletepost/' + post_id
+    And header token = auth_token
+    When method delete
+    Then status 200
+    And match response.status.avatar == 'Avatar Image Not Found'
+    And match response.status.postImg == 'Post Image Not Found'
+
+  Scenario: Try to delete a full user post
+    # Creating a avatar image
```
nathanielcaron, line 160: NIT: The endpoint POST `/postimages` creates an avatar and a post picture.
You could achieve this with a single call instead of two.

Automation/src/test/java/api/userPosts-delete.feature
```diff
@@ -53,3 +54,151 @@ Feature: Delete user post endpoint tests
...
+
+    # Create a post image
+    Given path 'image'
+    And header token = auth_token
+    And multipart file image = { read: '../data/post_image.jpeg', filename: 'image', contentType: 'image/jpeg' }
+    When method post
+    Then status 201
+    * def postImage_ID = response.id
+
+    # Creating a temporary user
```
nathanielcaron, line 176: NIT: The endpoint POST `/post` creates a user and a user post.
You could achieve this with a single call instead of two.

sahilsaini432: That's a good idea.
This way if the `post` endpoint gives an error this will also fail.
Would that be alright?

nathanielcaron: Yes, that's ok.

The endpoints you are currently calling could also fail.
A single call would just fail faster which is not a bad thing.

Service/api/api.js
```diff
@@ -425,6 +425,83 @@ exports.deleteImage = async (req, res) => {
     });
 };
 
+exports.deleteFullUserPost = async (req, res) => {
```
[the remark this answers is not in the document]
sahilsaini432: Ya, that would be good.
We would have to change some of our API tests to look for something else instead of `_id`'s then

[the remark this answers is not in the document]
sahilsaini432: @nathanielcaron Do you mind creating a new ticket for it?
I can pick it up alongside with `DBO-55`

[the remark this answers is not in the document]
nathanielcaron: Yep. We can discuss who can grab this one in standup
`DBO-60 Update backend endpoints and karate API tests to never return post Ids and access keys`

fredy20501 (approved):

Service/api/api.js
```diff
@@ -425,6 +425,83 @@ exports.deleteImage = async (req, res) => {
...
+  const userPostID = req.params.id;
+  const status = {};
+  let fullyDeleted = true;
+
+  if (!ObjectId.isValid(userPostID)) {
+    logger.info('Invalid Post ID');
+    return res.status(400).send({ message: 'Invalid Post ID Provided' });
```
nathanielcaron, line 435: NIT: Logger message and response message don't match

Service/api/api.js
```diff
@@ -425,6 +425,83 @@ exports.deleteImage = async (req, res) => {
...
+  }
+
+  // Deleting User Post
+  const post = await UTILS.deleteDBPost(userPostID);
+  if (post === UTILS.Result.NotFound) {
+    return res.status(404).send({ message: 'User Post Not Found' });
```
nathanielcaron, line 441: Missing logger info statement

Service/api/api.js
```diff
@@ -425,6 +425,83 @@ exports.deleteImage = async (req, res) => {
...
+  }
+  if (post === UTILS.Result.Error) {
+    return res.status(500).send({ message: INTERNAL_SERVER_ERROR_MSG });
```
nathanielcaron, line 444: Missing logger error statement

nathanielcaron (approved):

Service/api/api.js
sahilsaini432: The logger statement is added in the utils call. I can move it here if you want?

Service/api/api.js
sahilsaini432: Same for this one

Service/api/api.js
nathanielcaron: Ok either way is fine 👍

CarterMacLennan (approved):

JoaoPontes98 (approved):

PR #56, opened 2022-01-15
DBO-54 - Create UI Automation tests for the navbar component

nathanielcaron:
Added tests for navbar and terms and conditions components.


application/src/components/Delete.js
```diff
@@ -62,9 +62,9 @@ function Delete() {
 
   return (
     <>
-      <Container className="outer-container">
+      <Container className="outer-container" data-testid="delete-page">
```
CarterMacLennan, line 65: Could you clarify again what we decided for standardizing the class names & test ids?

CarterMacLennan (approved):

Automation/src/test/java/ui/termsConditions.feature
```diff
@@ -0,0 +1,51 @@
...
+    And match text('span[data-testid=privacy-policy]') == 'Nota.social Privacy Policy'
+
+  Scenario: Open terms and conditions after entering the site
+    # Enter website
+    Given driver frontendUrl
+    Then match enabled('button[data-testid=enter-btn]') == false
+    When click('input[data-testid=agree-checkbox]')
+    Then match enabled('button[data-testid=enter-btn]') == true
+    And click('button[data-testid=enter-btn]')
+    Then waitFor('nav[data-testid=navbar]')
```
fredy20501, line 33: I don't think we need to test that the button is disabled/enabled depending on the checkbox every time we enter the site. This will be used in almost all UI tests so if the button/checkbox stops working we wouldn't want all UI tests to fail, just the one that is specifically testing entering the website.

Automation/src/test/java/ui/termsConditions.feature
```diff
@@ -0,0 +1,51 @@
+@ui
+Feature: Terms and Conditions component automated tests
```
fredy20501, line 2: I think we should add a step that tests closing the modal (using the close button and maybe also the X of the modal - you'll have to use class selectors to find that one though).

application/src/components/Home.js
```diff
@@ -53,8 +53,8 @@ function Home() {
 
   return (
     <>
-      <Container className="home-page outer-container" data-testid="home-page">
-        <div className="h4 mb-0">Recent posts</div>
+      <Container className="home-page outer-container" id="home-page" data-testid="home-page">
```
fredy20501, line 56: Why did we add an `id` here when it already had a `data-testid`? (the `id` isn't used anywhere)

application/src/components/Delete.js
nathanielcaron: All components tested by the React component tests or the Karate UI automation tests will have a `data-testid`.
Only components that require an `id` for styling will have one.
Same thing for class names.

application/src/components/Home.js
nathanielcaron: Added by mistake.
Removed

Automation/src/test/java/ui/termsConditions.feature
nathanielcaron: Added steps to close using both buttons

Automation/src/test/java/ui/termsConditions.feature
nathanielcaron: Created a separate feature file for entering the website.
This could be expanded to include more detailed testing or used by other feature files to enter the website.

Removed these checks from other feature files.

Automation/src/test/java/ui/termsConditions.feature
```diff
@@ -23,12 +23,20 @@ Feature: Terms and Conditions component automated tests
...
 
+    # Close terms and conditions modal using close button
+    When click('button[data-testid=close-btn]')
+    Then retry(3, 1000).exists('div[data-testid=terms-conditions]') == false
+
+    # Close terms and conditions modal using X button
+    When click('a[data-testid=terms-conditions-link]')
+    Then waitFor('div[data-testid=terms-conditions]')
+    When click('button[class=btn-close]')
+    Then retry(3, 1000).exists('div[data-testid=terms-conditions]') == false
```
fredy20501, line 34: The class `btn-close` is a common one used in many bootstrap components which means it is very much possible that there could be multiple buttons with that class on the page at the same time. To prevent that from causing issues I suggest making your selector more precise (i.e., get the one that is inside the terms modal).

Example selector: (I think this would work)
```cucumber
When click('div[data-testid=terms-conditions] button[class=btn-close]')
```

nathanielcaron: Updated

Automation/src/test/java/ui/welcomePage.feature
```diff
@@ -0,0 +1,14 @@
+@ui
```
sahilsaini432, line 1: NIT: We can also add a test for clicking the "Enter Site" button without accepting terms and conditions and making sure it doesn't render the homepage.

nathanielcaron: This is outside the scope of this ticket which was only for the navbar.

This welcome page file will likely be used by other feature files to enter the site. We can create a separate one later to test all functionality of this page.

sahilsaini432 (commented):
It might be just on my side but when I try to run the UI tests using this `mvn test -Dtest=UiTests -Dkarate.env="dev"` it getting stuck in a loop.

This is the message that it shows
Screen Shot 2022-01-16 at 3 26 54 PM

nathanielcaron:
> It might be just on my side but when I try to run the UI tests using this `mvn test -Dtest=UiTests -Dkarate.env="dev"` it getting stuck in a loop.
> 
> This is the message that it shows Screen Shot 2022-01-16 at 3 26 54 PM

Yes I am getting the same issue on my M1 macbook air with java 17.

The tests run and succeed in Intellij and VS Code.

I create a ticket to investigate this but I'm pretty sure this is because it should be run using java 8.

fredy20501 (approved):

JoaoPontes98 (approved):

sahilsaini432 (approved):

PR #58, opened 2022-01-16
DBO-55 : Update karate API tests for get user posts endpoints, where a user in the database is used, to create temporary users instead

sahilsaini432:


Automation/src/test/java/utils/global_user_post.feature
```diff
@@ -0,0 +1,15 @@
+@ignore
+Feature: Create Post
```
fredy20501, line 2: I think this file name (`global_user_post.feature`) should be changed since it doesn't really describe what it does. I think something like `createPost.feature` would be more clear. (We may also want to rename `cleanup.feature` to `deletePost.feature`)

fredy20501:
In `userPosts-post.feature` I had commented out lines which used to make tests fail since the posts were using  hardcoded userIds which made the test failed if they weren't in the DB. You can find them by looking for these comments:
```cucumber
# TODO: update this with DBO-55
```
We should fix this as part of this PR since it is relevant. I suggest creating a user for them (calling POST `/user`) and using the returned userID to override the one in the test data before creating the post.

sahilsaini432:
> In `userPosts-post.feature` I had commented out lines which used to make tests fail since the posts were using hardcoded userIds which made the test failed if they weren't in the DB. You can find them by looking for these comments:
> 
> ```
> # TODO: update this with DBO-55
> ```
> 
> We should fix this as part of this PR since it is relevant. I suggest creating a user for them (calling POST `/user`) and using the returned userID to override the one in the test data before creating the post.

That is something I wanted to talk about in the next meeting.
I think we should remove the check for users from these cases. In this case, since the tests are for the `userpost` endpoint, to create a user with it I would have manually provide an `author` id when I call the endpoint. But, since we are planning on removing the `id`'s from the create post and user endpoints I won't be able to manually assign it to a post.
Any suggestions on what can we do to solve this?

nathanielcaron:
> > In `userPosts-post.feature` I had commented out lines which used to make tests fail since the posts were using hardcoded userIds which made the test failed if they weren't in the DB. You can find them by looking for these comments:
> > ```
> > # TODO: update this with DBO-55
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We should fix this as part of this PR since it is relevant. I suggest creating a user for them (calling POST `/user`) and using the returned userID to override the one in the test data before creating the post.
> 
> That is something I wanted to talk about in the next meeting. I think we should remove the check for users from these cases. In this case, since the tests are for the `userpost` endpoint, to create a user with it I would have manually provide an `author` id when I call the endpoint. But, since we are planning on removing the `id`'s from the create post and user endpoints I won't be able to manually assign it to a post. Any suggestions on what can we do to solve this?

I don't think we have to remove the ids from endpoints creating users. The endpoint to delete users will never be exposed in production anyways.

fredy20501 (approved):

nathanielcaron (approved):

nathanielcaron:
Great work. The `@ignore` tag is very useful for feature files only used by other tests.

JoaoPontes98 (approved):

CarterMacLennan (approved):

PR #59, opened 2022-01-18
DBO-51: Karate pagination tests + case insensitive tag search

JoaoPontes98:
### The pagination tests:
- I added a parameter to `/userposts` + `/recentposts` for a post limit per page.
- The tests create new users + posts and test if the parameter works by limiting the number of posts being returned.

### Case insensitive tag search
- I updated the schema to use mongoose's baked-in functionality to convert tags to lowercase.
- I updated the `/userposts` to first convert the request tags to lowercase, and then proceed as before.
- I changed the `userPosts-post.feature` tests to match for lowercase tags.


Automation/src/test/java/api/pagination.feature
```diff
@@ -0,0 +1,127 @@
...
+
+    # Create 2 Temporary User Posts
+    Given path 'userpost'
+    And header token = auth_token
+    And request userpostJSON
+    When method post
+    Then status 201
+    * def post1_access_key = response.post.access_key
+    * def post1_id = response.post._id
+
```
sahilsaini432, line 27: You might wanna change this part to use `/post` endpoint to create post and user at the same time.
Or I can change it in `DBO-60 Update backend endpoints and karate API tests to never return post Ids and access keys`. This will fail afterwards because the `user` endpoint will not return the id.

Automation/src/test/java/api/pagination.feature
```diff
@@ -0,0 +1,127 @@
...
+    Given path 'userpost'
+    And header token = auth_token
+    And request userpostJSON
+    When method post
+    Then status 201
+    * def post2_access_key = response.post.access_key
+    * def post2_id = response.post._id
+
```
sahilsaini432, line 35: Same for this one, `_id` from `/userpost` endpoint will soon be hidden.

Automation/src/test/java/api/pagination.feature
```diff
@@ -0,0 +1,127 @@
...
+    And header token = auth_token
+    When method delete
+    Then status 200
+
+    # Delete User
+    Given path 'user/' + user_id
+    And header token = auth_token
+    When method delete
+    Then status 200
+
```
sahilsaini432, line 68: NIT: Here you can also use the `/deletepost` endpoint so it deletes the user and the post at the same time.

Automation/src/test/java/api/pagination.feature
```diff
@@ -0,0 +1,127 @@
...
+    Scenario: Test Pagination of userPosts/
```
sahilsaini432, line 69: Same Suggestions for this scenario

Automation/src/test/java/api/pagination.feature
JoaoPontes98: Good catch Sahil! should be fixed now

Service/api/api.js
```diff
@@ -251,7 +251,10 @@ exports.getUserPosts = (req, res) => {
   // Get current page number
   const pageNumber = req.body.page ? (req.body.page - 1) : 0;
 
-  UserPost.aggregate(searchFilters).skip(pageNumber * POST_LIMIT).limit(POST_LIMIT)
+  // Get # of posts per page
+  const postLimit = req.body.post_limit ? req.body.post_limit : POST_LIMIT_DEFAULT;
```
fredy20501, line 255: We should update the raml example to include the new `post_limit` attribute.

Service/api/api.js
```diff
@@ -280,7 +283,10 @@ exports.getRecentPosts = (req, res) => {
     ];
   }
 
-  UserPost.aggregate(searchFilters).limit(POST_LIMIT)
+  // Get # of posts per page
+  const postLimit = req.body.post_limit ? req.body.post_limit : POST_LIMIT_DEFAULT;
```
fredy20501, line 287: (Again) We should update the raml example to include the new `post_limit` attribute.

Service/api/api.js
fredy20501, line 255: Nit: We can use the [Nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) to simplify this operation (this is totally optional, just a useful thing to know)
```js
const postLimit = req.body.post_limit ?? POST_LIMIT_DEFAULT;
```

fredy20501 (approved):
Looks good!

Service/api/api.js
```diff
@@ -183,13 +183,13 @@ exports.getUserPosts = (req, res) => {
 
   // Check if the filters have tags
   if (req.body.tags?.length > 0) {
+    providedTags = req.body.tags.map((tag) => tag.toLowerCase());
```
nathanielcaron, line 186: NIT: This could be made into a small function and unit tested. It could be done as part of `DBO-59 Look into creating more unit tests for the service`

nathanielcaron (approved):

nathanielcaron:
Good stuff 🥇

sahilsaini432 (approved):
Nice Work 💯

CarterMacLennan (approved):

PR #60, opened 2022-01-22
DBO 60 - Update backend endpoints and karate API tests to never return post Ids and access keys

sahilsaini432:


Automation/src/test/java/utils/getPost.feature
```diff
@@ -0,0 +1,15 @@
+@ignore
+Feature: Delete Post
```
nathanielcaron, line 2: Should probably say `Feature: Get Post`

Service/api/api.js
```diff
@@ -83,7 +83,6 @@ exports.createUserPost = (req, res) => {
 
     return res.status(201).json({
       post: {
-        _id: newUserPost._id,
```
nathanielcaron, line 86: NIT: Might want to consider also having a `.select('-_id')` in the create post endpoint for consistency.
Since we don't return it, we should not request it in the first place.

nathanielcaron (approved):

Service/api/api.js
sahilsaini432: I don't think we can hide it here. Since we not actually calling any mongo's filter/search function I can't add the `.select('-_id')` tag here. (I tried it too 😅)

JoaoPontes98 (approved):

fredy20501 (commented):
Great work!
Could you add tests that actively check that `_id` and `access_key` are **not** returned from the APIs you modified? That way if someone accidentally re-adds the ids we'll know since tests will fail.
(Ideally, we should make new/separate tests for this to keep tests as atomic as possible but you could just add the checks for `_id` and `access_key` to a few of the existing ones too)

nathanielcaron:
> Great work! Could you add tests that actively check that `_id` and `access_key` are **not** returned from the APIs you modified? That way if someone accidentally re-adds the ids we'll know since tests will fail. (Ideally, we should make new/separate tests for this to keep tests as atomic as possible but you could just add the checks for `_id` and `access_key` to a few of the existing ones too)

Good idea.
I would add separate scenarios for this.

Automation/src/test/java/api/id-accessKey.feature
```diff
@@ -0,0 +1,95 @@
...
+    * def post_id = post1.response._id
+
+    # Call the PATCH '/userpost' endpoint
+    Given path 'userpost/' + post_id
+    And header token = auth_token
+    And request { update: { title: 'new title' } }
+    When method patch
+    Then status 200
+    * match each response..['access_key'] == '#notnull'
+    * match each response..['_id'] == '#notpresent'
```
fredy20501, line 38: Why are you using `#notnull` for the `access_key` and `#notpresent` only for the `_id`? Shouldn't they both be not present?

sahilsaini432: Oh, good catch. I will change it

fredy20501 (approved):

CarterMacLennan (approved):

PR #61, opened 2022-01-23
DBO 66 - Update Post component so that the image can be expanded by clicking on it

sahilsaini432:
Changes Made to the post Component

ezgif com-gif-maker


nathanielcaron:
Could you include a short description of the changes made and an image or gif?

nathanielcaron:
I think that when we expand these images it should not say 'post image' and 'avatar image' in the header.
I would suggest displaying the post title for the post image, and the user's name for the avatar image.
Or we could also not have a header for the modal.

nathanielcaron:
I ❤️ the look of this on desktop but the mobile view really suffers because of these changes...

If we want to keep this layout we'll have to set a minimum size for the map and set the zoom to be closer.

Screen Shot 2022-01-27 at 10 14 35 PM

application/src/components/post/Post.js
```diff
@@ -22,37 +26,59 @@ function Post({postData}) {
...
-              <div className="text-muted text-center">{postData.location_string}</div>
-            </FCol>
-          </FRow>
-          <div>
-            <img src={postData.img_URL} />
-          </div>
-        </FCol>
-      </FRow>
-    </Container>
+      <Modal size='xl' scrollable show={avatarImageModal} onHide={() => setAvatarImageModal(false)}>
```
CarterMacLennan, line 75: With the static size, I just wanted to double check that this fits all reasonable screen sizes?

sahilsaini432: I think the size here is decided dynamically according to the screen size since it's a bootstrap component. 
https://react-bootstrap.netlify.app/components/modal/#modals
(If you wanna take a more detailed look)

nathanielcaron (approved):

CarterMacLennan (approved):

fredy20501 (approved):

JoaoPontes98 (approved):

PR #62, opened 2022-01-23
DBO 57 - Call the exposed delete user post endpoint from the frontend when deleting a user post

sahilsaini432:


application/src/api/api.js
```diff
@@ -43,21 +43,16 @@ export const getPostByAccessKey = async (accessKey) => {
...
-  //   return response.data;
-  // });
-  
-  //! Temporary data
-  _id;
-  return { message: 'User Post Deleted Successfully' };
+  return await requestWithToken(async() => {
+    const response = await axios({
+        method: 'DELETE',
+        url: serviceUrl + '/deletepost/' + _id,
```
nathanielcaron, line 49: I didn't realize that the endpoint route was DELETE `/deletepost`.

We should rename it to DELETE `/post` (The method should not be in the route, that's redundant).
It will also make consistent with the exposed endpoint to create a user post and user.

nathanielcaron (approved):

fredy20501 (approved):

JoaoPontes98 (approved):

CarterMacLennan (approved):

PR #65, opened 2022-01-25
CD Pipeline

CarterMacLennan:
casual 130+ commits...... (80% of them is just triggering the pipeline)


Service/db/init-mongo.js
```diff
@@ -0,0 +1,12 @@
+db.createUser(
```
nathanielcaron, line 1: What is this?

CarterMacLennan: ahh thx u, I will remove. I was looking for that actually but I thought I had removed it already.

nathanielcaron:
Note: Once we merge this in, we'll have to update our local `.env` files to use port number 3001 for the service.

application/Dockerfile.dev
```diff
@@ -0,0 +1,17 @@
+# Specify a base image
+FROM node:14.17.0-alpine as alpine
+
```
fredy20501, line 3: Are we doing anything with this `Dockerfile.dev` file? If not we might want to delete this file.

Service/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+WORKDIR /app
+
+COPY package.json .
+RUN npm install
+COPY . .
+
+EXPOSE 3001
+
+CMD ["npm","run","dev"]
```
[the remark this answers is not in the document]
nathanielcaron: Yes.
Once again, this package must be installed globally `npm install pm2 -g`.
See: https://www.npmjs.com/package/pm2

nginx/configs/default.conf
```diff
@@ -0,0 +1,20 @@
+upstream api {
+    server api:3001;
+}
+
+upstream client {
+    server client:3000;
+}
+
+server {
+    listen 80;
```
sahilsaini432, line 10: Is this port 80 here for the EC2 Container?

nathanielcaron: That is what Nginx uses

"Port 80 is the port number assigned to commonly used internet communication protocol, Hypertext Transfer Protocol (HTTP)"

fredy20501 (approved):
Great work Carter! 🥇

sahilsaini432 (approved):
Nice Work Carter 💯

JoaoPontes98 (approved):

nathanielcaron (approved):

nathanielcaron:
Good stuff 💯

PR #67, opened 2022-01-27
DBO-13: Implement the Create Page

fredy20501:
# Description
- Implemented the Create Page.
- I mean there's just too much stuff in this PR to describe. Sorry guys ¯\\\_(ツ)\_/¯
- Have fun reviewing! (don't hesitate to add comments/questions! 😄)

# Demo


application/src/components/CopyButton.js
```diff
@@ -0,0 +1,62 @@
...
+      .then(() => {
+        // Show feedback on success
+        setIsCopied(true);
+        // Hide it after a few seconds
+        setTimeout(() => {
+          setIsCopied(false);
+        }, 1500);
+      })
+      .catch((err) => {
+        logger.error(err);
```
nathanielcaron, line 35: Should probably use `logger.error(err.message);`

application/src/components/Create.js
```diff
@@ -1,14 +1,420 @@
...
-      <div data-testid="create-title" className="h4 mb-0">Follow these steps to create a new post</div>
+    <Container className="outer-container">
+      <div className="h4"><Number num={props.num}/> {props.title}</div>
+      {props.children}
     </Container>
+  );
+}
+Section.propTypes = {
+  num: PropTypes.string.isRequired,
+  title: PropTypes.node.isRequired,
```
nathanielcaron, line 43: Would titles be anything else than strings?

application/src/components/CopyButton.js
fredy20501: I don't know if that error has a message attribute or not. That's why I didn't put it.

application/src/components/Create.js
```diff
@@ -1,14 +1,420 @@
...
+      setIsCreateError(true);
+
+      // Delete images (if any)
+      if (avatarId) deleteImage(avatarId);
+      if (pictureId) deleteImage(pictureId);
+    }
+  }
+
+  const sendEmail = () => {
+    // Todo: implement email sending in the backend
```
nathanielcaron, line 161: `[1] DBO-73 Call the send email endpoint from the create post component`

application/src/components/Create.js
fredy20501: Currently no, but I like keeping things flexible in general. I think originally I was using `div` and stuff which is why I had it set like this but the format changed as I kept working on it so that's no longer the case.

application/src/components/Create.js
```diff
@@ -1,14 +1,420 @@
...
+    setCreated(false);
+    setAccessKey('');
+    setIsCreateError(false);
+    setEmail('');
+  }
+
+  return (
+    <>
+      <Container className="outer-container" data-testid="create-page">
+        <div className="h4 mb-0">Follow these steps to create a new post</div>
```
nathanielcaron, line 192: NIT: `<div className="h4 mb-0">Follow the steps below to create a new post</div>`

application/src/components/Create.js
```diff
@@ -1,14 +1,420 @@
...
+              <br/>
+              <div style={{width:'100%', height:'350px'}}>
+                <LocationPickerMap onPositionChange={setPosition} setIsTruePosition={setIsTruePosition} />
+              </div>
+              <Alert
+                variant="danger"
+                className="mb-0 mt-3"
+                hidden={position !== null}
+              >
+                <MdErrorOutline/> You must select a location.
```
nathanielcaron, line 210: Should there be a closing tag after the text?
If we use the self-closing tag then I think the text should be on a new line.

application/src/components/Create.js
```diff
@@ -1,14 +1,420 @@
...
+              </Alert>
+            </Section>
+
+            <Section num="2" title="Avatar">
+              <div>Tell everyone who you are!</div>
```
nathanielcaron, line 215: NIT: <div>Tell us more about you!</div>

application/src/components/Create.js
```diff
@@ -1,14 +1,420 @@
...
+                        <AvatarUploadButton setAvatarImg={setAvatarImg}/>
+                      </Form.Group>
+                    </Col>
+                  </Row>
+                </Col>
+              </Row>
+            </Section>
+
+            <Section num="3" title="Content">
+              <div>What do you want to share?</div>
```
nathanielcaron, line 251: NIT:  `<div>What would you like to share?</div>`

application/src/components/maps/LocationPickerMap.js
```diff
@@ -47,42 +50,24 @@ function LocationPickerMap(props) {
...
-    controlUI.style.boxShadow = '0 2px 6px rgba(0,0,0,.3)';
-    controlUI.style.cursor = 'pointer';
-    controlUI.style.marginTop = '8px';
-    controlUI.style.marginBottom = '22px';
-    controlUI.style.textAlign = 'center';
-    controlUI.title = 'Click to use your current location';
-    controlDiv.appendChild(controlUI);
+    const controlUI = 
+      <Button 
+        title="Click to use your current location"
```
nathanielcaron, line 63: NIT: `title="Use your current location"`

nathanielcaron:
Personally, I would vote for keeping the email address when storing users.
I appreciate the fact that we don't currently use it but It doesn't really make sense to have a database object with only two attributes.
It would allow for more flexibility in case we'd like to add features or restrictions for users.
We can discuss this during standup.

nathanielcaron:
Outstanding work @fredy20501 🚀 
This definitely deserves the PR of the year award 🏆

application/src/components/Create.js
fredy20501: So the `<MdErrorOutline/>` is an icon, which is why it is self-closing. This is an icon with text next to it, basically. I'm not sure I understand what you're suggesting.

nathanielcaron: Makes sense. It just looked weird to me to have the tag and the text on the same line.
...
`<MdErrorOutline/>`
`You must select a location.`
...

fredy20501: Ah, right. I put it on the same line since that's how it looks on the page.

application/src/components/Create.js
```diff
@@ -1,14 +1,420 @@
...
+                hidden={isPostValid()}
+              >
+                <MdErrorOutline/> Some required fields are missing values.
+              </Alert>
+            </Section>
+
+            <When condition={isPostValid()}>
+              <Post postData={getPostData()}/>
+
+              <Section num="7" title="Publish">
```
sahilsaini432, line 335: Question: Shouldn't we display this section even when the post is not valid?

application/src/components/Create.js
```diff
@@ -1,14 +1,420 @@
...
+  const preventSubmit = (event) => {
+    // Prevent default form submission
+    event.preventDefault();
+    event.stopPropagation();
+  }
+
+  const openModal = () => {
+    setShowConfirmationModal(true);
+    // Reset create error if was set
+    setIsCreateError(false);
```
sahilsaini432, line 122: Question: What is this useState checking for?

application/src/components/CopyButton.js
```diff
@@ -0,0 +1,62 @@
...
+  const [isCopied, setIsCopied] = useState(false);
+
+  async function copyTextToClipboard(text) {
+    // Try to use clipboard API
+    if ('clipboard' in navigator) {
+      return await navigator.clipboard.writeText(text);
+    }
+    else {
+      // Use this as last resort (deprecated)
+      return document.execCommand('copy', true, text);
```
sahilsaini432, line 20: Question: If `document.execCommand` is deprecated, is there an advantage of using it here?

application/src/components/Create.js
fredy20501: Also if we put it on a separate line we would need to manually add a non-breaking space `&nbsp;` like this:
```xml
<MdErrorOutline/>&nbsp;
You must select a location.
```

application/src/components/Create.js
fredy20501: When the post is not valid I don't display either the preview or the publish section. The preview because it can look broken if things are missing. The publish section because they shouldn't be able to submit if invalid and I want the error message to be easy to see, so this way the error message is at the bottom of the screen (and there isn't anything else past it).

I hope that made sense 😅

application/src/components/Create.js
fredy20501: The `isCreateError` is a flag variable that determines whether we show the "Post could not be created" message at the very bottom of the screen. It is set to true when the post fails to be created (so the message appears), and reset whenever you open the confirmation modal (so that if it fails again the message will re-appear so it is clear that it failed again)

application/src/components/CopyButton.js
fredy20501: This is basically to support old browsers in case they don't implement the `clipboard` API. At this point, all modern browsers have that API so we will rarely get to this else statement. The fact that it is deprecated doesn't mean that it won't work, it just means that modern browsers don't need to support it anymore.

sahilsaini432 (approved):
Nice work Fred 💯 
Works Great 👍

CarterMacLennan (approved):

JoaoPontes98 (approved):

nathanielcaron (approved):

PR #68, opened 2022-01-28
DBO-74: Implement Reverse Geocoding when Creating Posts

fredy20501:
## Summary
- Created a function to call the Geocoding Google API so we can get the location string when creating a post.
  - The geocoding updates on a timer to prevent users from spamming requests and reaching the max request per second.
- Changed backend max image size to 15MB to match the frontend.
- Added the location string as the title of the map modal.

## Notes
- This PR is set to merge into #67 since it is based on the create page.
- It is really fun to play around clicking in random countries. 😄


application/src/components/Create.js
```diff
@@ -18,6 +18,7 @@ import { TermsLink, TermsCheckbox } from './terms/Terms';
 
 // API
```
fredy20501, line 19: Don't merge this branch until it gets 4 reviews plz.

fredy20501:
Don't merge this until it gets 4 reviews plz. I know the big green button is tempting but it is _forbidden_.

image

application/src/components/maps/Geocoder.js
```diff
@@ -0,0 +1,75 @@
...
+    logger.error('Geocoding error: ' + e.message)
+    callback(DEFAULT_LOCATION_STRING);
+  });
+}
+
+const extractGeocodeResult = (results) => {
+  // Extract the country, province, city (if possible)
+  let country, province, city;
+  const foundAll = () => country && province && city;
+  for (let i=results.length-1; i>=0 && !foundAll(); i--) {
```
sahilsaini432, line 40: Question: What is the use of the `foundAll()` constant here?

fredy20501: So you can see it defined on the line above. It is basically a shorthand for checking `country && province && city` (i.e., whether those variables have been set). I'm using it to make the code more readable, so I don't have to write `country && province && city` in both loops.

The actual reason why I'm checking `foundAll()` in the first place is so that the loops stop once we find the 3 values (because the Geocoding response can sometimes have upwards of a dozen results and we usually don't need to check all of them).

CarterMacLennan (approved):
Great work Fred, looks good to me.

sahilsaini432 (approved):
Works Great Fred 💯

JoaoPontes98 (approved):

nathanielcaron (approved):

fredy20501:
Thanks guys! I'll merge this in after #67

PR #69, opened 2022-01-29
DBO-28: Implement the Search Page

fredy20501:
# Summary of Changes
## Search Page
- The search page is fully functional
- Notes:
  - Search vs Refresh:
    - Search button triggers a new search using the currently selected tags
    - Refresh button re-does the last request (regardless of what tags are currently selected)
  - Updated `getUserPost` endpoint to also return the total number of posts matched (before skip/limit).
    - This is used in the Search page to a) display the number of results and b) know when to stop showing the "Load More" button.
    - Updated Raml & API tests accordingly

## Other Stuff
- The tags are now capitalized automagically using the power of CSS `text-transform`.
- Added error checking for avatar and picture in the Post component.
- Update some in-line CSS to use Bootstrap's pre-defined classes.

# Demo


fredy20501:
Don't merge this until it gets 4 reviews, please. 🛑 
image

Service/api/api.js
```diff
@@ -240,35 +240,54 @@ exports.getUserPosts = (req, res) => {
...
+    {
+      $facet: {
+        // Skip & limit posts returned
+        posts: [
+          { $skip: pageNumber * postLimit },
+          { $limit: postLimit }
+        ],
+        // Give back total number of posts matched (before skip/limit)
+        info: [
+          { $count: 'totalCount' }
```
sahilsaini432, line 280: Don't we need to define `totalCount` first?
Or is it like a special keyword in MongoDB?

application/src/components/post/Post.js
```diff
@@ -8,10 +9,21 @@ import './post.css';
...
 import PlaceholderAvatar from '../../resources/images/placeholder-avatar.png';
 
+// Show placeholder image if avatar image fails
+function handleAvatarImgError(e) {
+  if (e.target.src != PlaceholderAvatar) e.target.src = PlaceholderAvatar;
+}
+
 function Post({postData}) {
+  const [imgURL, setimgURL] = useState(postData.img_url);
+
```
sahilsaini432, line 22: NIT: You can also do a similar thing for the avatar img?

sahilsaini432 (approved):
Nice Work Fred 💯

Service/api/api.js
fredy20501: `totalCount` is the name of the variable that will hold the count value in the returned results. Here's an example result:
```js
[{
  posts: [
    // Matched posts are here
  ],
  info: [
    totalCount: 4
  ]
}]
```

application/src/components/post/Post.js
fredy20501: This was needed because I needed a way to hide the image on error, which I'm doing by setting the image URL to null since there is a `hidden={}` check for the image container in the code below, and we can't modify the `img_url` prop directly. It was not needed for the avatar because for the avatar if it fails I just change the URL to the placeholder one (see `handleAvatarImgError()` above).

application/src/components/Search.js
```diff
@@ -1,13 +1,197 @@
...
+
+  const refreshPosts = () => {
+    // Call getPost even if the page hasn't changed
+    if (page === 1) getPosts();
+    // Else reset the page (which will trigger getPosts)
+    else setPage(1);
+  }
+
+  const loadMorePosts = () => {
+    setPage(page+1);
```
nathanielcaron, line 46: Should we not set a max page number?

It could be quite high so users rarely reach it, something like 50.

nathanielcaron (approved):

application/src/components/Search.js
fredy20501: Well, I loaded about 125 posts on the same page and didn't have any issues, so I don't think we need a max page number. Adding a max page number would limit the user experience and we don't even know if it's needed. I say we hold off until we know for sure if no limit is a problem or not.

CarterMacLennan (approved):
Looks great to me :)

JoaoPontes98 (approved):

PR #70, opened 2022-01-29
DBO-42 - Create loading spinner component for frontend

nathanielcaron:
- Created simple loading spinner component
- Added fade-in example for components in Home page (can add to other pages later)
- Made background color match for scroll-pull

dbo-42


application/src/components/Home.js
```diff
@@ -42,21 +46,24 @@ function Home() {
...
-        <div data-testid="home-title" className="h4 mb-0">Recent posts</div>
-      </Container>
-      <Post postData={postData}/>
+      <If condition={isLoading}>
+        <Then>
+        <Container className="home-page outer-container" data-testid="home-page">
+          <LoadingSpinner message="Looking for recent posts..." size="10rem"/>
+        </Container>
+        </Then>
+      </If>
```
fredy20501, line 57: If you're just using `If`-`Then` without the `Else`, you can use the `When` component instead:
```html
<When condition={isLoading}>
   ...
</When>
```

nathanielcaron: Updated

sahilsaini432 (approved):
Nice work 💯 
I like the fade effect before the post appears 👍

fredy20501 (approved):
Looks good! 👍 
- I don't know how we'll want to merge this with my work in PR #71 since I also added a spinner when loading the posts.
- For the fading, we could also do it in `MainPage.js` and do a fade in/out for each page as part of the `Switch` case.

nathanielcaron:
> Looks good! 👍
> 
> * I don't know how we'll want to merge this with my work in PR #71 since I also added a spinner when loading the posts.
> * For the fading, we could also do it in `MainPage.js` and do a fade in/out for each page as part of the `Switch` case.

We can remove the changes in `Home.js` from this PR and merge in the spinner component.
Once it's merged, you'll be able to bring in the changes and add back the fade in and use this spinner instead.

JoaoPontes98 (approved):

PR #71, opened 2022-01-30
DBO-30: Implement the Home Page

fredy20501:
## Description
- The home page is fully functional.
  - Works very similarly to the Search page.
- Updated `getRecentPosts` endpoint to also return the total number of posts matched (before skip/limit).
  - This is used in the Home page to know when to stop showing the "Load More" button.
  - Updated Raml & API tests accordingly.
- Added nota logo as the favicon.

Note: this PR is set to merge into #69


sahilsaini432 (approved):
Nice Work Fred 💯

CarterMacLennan (approved):
u r v. awesome Fred

application/src/components/Home.js
```diff
@@ -1,62 +1,141 @@
...
+  const addPosts = (newPosts) => {
+    setPosts(posts.concat(newPosts));
+  }
+
+  // Get recent posts on page load & whenever lastDate changes
+  useEffect(getPosts, [lastDate]);
+
+  const LOAD_DURATION = 800;
+  // Set the isLoading state for the given duration
+  const fakeLoading = (setLoad) => {
```
nathanielcaron, line 74: Personally, I don't really see the need for a randomized loading.

A base loading time, like 500ms, plus whatever querying the backend for posts should be ok.
It will vary depending on the number of posts in the database anyways.

application/src/components/Home.js
```diff
@@ -1,62 +1,141 @@
...
+              className="clickable hover-outline rounded h2 mb-0" 
+              style={{color:'var(--bs-primary)'}} 
+              onClick={refreshPosts}
+            />
+          </FCol>
+        </FRow>
       </Container>
-      <Post postData={postData}/>
+
+      <When condition={showResults && (!isLoading || isPaginating)}>
```
nathanielcaron, line 104: Should be able to easily replace the <When> with a <Fade> for the posts here and one for the error message below.

application/src/components/Home.js
fredy20501: The way it is set up is that the delay is a _minimum_ delay, not a base loading time. It is set up so that I have two flags (`isLoading` and `showResults`) and the posts are only shown once both flags are set. So if the API call is faster, it will always show the posts after the fixed delay of 800ms, which is why I added a bit of randomness so it isn't always constant. But I guess I'm not opposed to keeping it constant

application/src/components/Home.js
fredy20501: Actually, I started working on that and just replacing the `<When>` will make the first posts fade in, but not when you load more since that just appends more posts to a div that is already shown. I'll have to figure out some trickery to make it work when loading more.

nathanielcaron: Ok, if it's too complicated, we could simply have the fade in when first loading the page with the first set of posts.
Loading more posts could just keep being appended and not fade in.

JoaoPontes98 (approved):

application/src/components/Home.js
fredy20501: Yeah, I'll look into it a bit and we'll see if I can figure something out quickly. Otherwise I'll keep it basic and move on.

fredy20501: @nathanielcaron 
I just pushed a final commit which merges your spinner and fade animations (used in both Home and Search). You can just review that last commit if you have already reviewed everything else.

I had to add a new package (`react-transition-group`) to make the "Load More" work when adding new posts. (technically this is not a new package since `react-bootstrap` uses that package internally for its transitions)

nathanielcaron (approved):

PR #72, opened 2022-01-30
DBO-61: Email access key api endpoint

JoaoPontes98:
**Changes made**
1. Renamed `s3.js` to `aws.js` and adjusted the affected files accordingly.
2. Added a method to `aws.js` called `sendEmail`. It takes in an email object as a parameter and returns a `.promise()`.
3. Added an api endpoint called `/sendakemail` which sends an email with an access-key. The endpoint takes `to_email` and `access_key` as parameters.
4. Added the new email enpoint to the raml documentation
5. I made our own noreply gmail account. I'll share the credentials with yall privately.

**Notes**
This endpoint will not work unless the address you're sending to is a verified user in our AWS account. This is because our account is still in sandbox mode. I'm currently in the process of getting us switched to production mode, which will allow us to send emails to non-verified users. In the meantime, if you wish to test the endpoint, here's how you add your email to our verified list

1. Log into your IAM credentials and navigate to the `Amazon Simple Email Service`.
2. On the left side select `Verified identities`
3. Click `Create Identity` and proceed with the creation process.

**Screen Shots**
Below is an example of an email sent by the api endpoint. The formatting of the email body is done directly in the endpoint by using html.
image


Service/api/api.js
```diff
@@ -638,3 +638,26 @@ exports.createFullUserPost = async (req, res) => {
...
+
+  const email = {
+    from: 'notasocial.noreply@gmail.com',
+    to: req.body.to_email,
+    subject: 'Nota Post Access-Key',
+    html: `<p>The access-key to your new Nota post is <strong>${req.body.access_key}</strong>`
+        + '<p><div style="color:red;">WARNING:</div>'
+        + 'If this access-key is lost, you will no longer be able to delete the Nota post.</p>'
+        + '<p>Happy adventuring!</p>'
+  };
```
fredy20501, line 655: Nit: On the frontend pages we actually called the `access_key` as an "access code" so we should also call it an access code here for consistency.

Service/api/api.js
```diff
@@ -638,3 +638,26 @@ exports.createFullUserPost = async (req, res) => {
     });
   });
 };
+
+exports.sendAKEmail = async (req, res) => {
+  if (!req.body.access_key || !req.body.to_email) {
+    return res.status(400).send({ message: INVALID_REQUEST_ERROR_MSG });
+  }
+
+  const email = {
```
fredy20501, line 647: We might want to add a bit more info in the email body/subject so the user knows which post this relates to. Maybe add the author name & post title in the email?

Service/api/api.js
nathanielcaron: There should also be a disclaimer telling users not to share the access code.

Service/api/api.js
nathanielcaron: I agree the post title and author name should be included.
The front end will have that information ready.

Service/routes/routes.js
```diff
@@ -134,4 +134,10 @@ ROUTER.delete(
   USE(API.deleteFullUserPost)
 );
 
+ROUTER.post(
+  '/sendakemail',
```
nathanielcaron, line 138: NIT: could rename path to `/akemail`

nathanielcaron:
Awesome work @JoaoPontes98 🚀

nathanielcaron (approved):

fredy20501 (approved):
Nice work! 🎆

CarterMacLennan (approved):

PR #81, opened 2022-02-18
DBO-46: Implement captcha puzzle for creating posts

fredy20501:
vvvvvvvvvvvvvv
vvvvvvvvvvvvvv
**IMPORTANT**: you don't need to review the code under `application/src/components/captcha` since that is the content of an npm package.
^^^^^^^^^^
^^^^^^^^^^

## Summary of Work
- Added a captcha slider puzzle to the create page.
- Used the package [@slider-captcha](https://www.npmjs.com/package/@slider-captcha/react) 
  - Note: the react part of that package was set to React 16.8 so it didn't directly work with our project (we use React 17). Instead I manually copied the parts I needed into our frontend as if it was our own component (under `application/src/components/captcha`). I also made modifications and simplifications to it so it technically isn't a direct copy-paste. The backend ("core") part was installed as an npm package normally.
- Implemented [`express-session`](https://www.npmjs.com/package/express-session) in the backend.
  - This was needed for the captcha verification process (see the section below).
  - Session data is stored in our MongoDB database under the collection named "sessions".
  - Note: a new environment variable is used for the backend called `SESSION_SECRET`. It can be a string of alphanumeric characters of any length.

## How it Works
1. The frontend sends a request to the `/captcha/create` API which creates a new puzzle.
2. The API sends back the puzzle image and stores the solution in a session (using `express-session`).
3. Once the user releases the slider, the frontend sends the current distance of the slider to the `/captcha/verify` API.
4. The API verifies the solution against what was stored in the session and returns whether it is valid (within a certain tolerance). If valid, it also returns a token (which is also stored in the session).
5. If the frontend receives a fail, it automatically creates a new puzzle (back to step 1). If it is successful, the frontend keeps the token and attaches it to its request to the create post API.
6. The create post API checks the token against the one stored in the session. If it doesn't match, request fails with status 403. If it matches the post is created.
7. If the frontend receives status 403 it resets the puzzle and asks the user to do it again.

Note: the verification of the token is only done for the production endpoint since the API tests can't handle the captcha verification process.


Service/api/api.js
```diff
@@ -730,3 +733,42 @@ exports.sendAKEmail = async (req, res) => {
...
+      logger.error(err.message);
+      return res.status(500).send({ message: INTERNAL_SERVER_ERROR_MSG });
+    });
+};
+
+exports.verifyCaptcha = async (req, res) => {
+  sliderCaptcha.verify(req.session.captcha, req.body)
+    .then(function (verification) {
+      if (verification.result === 'success') {
+        req.session.captchaToken = verification.token;
```
nathanielcaron, line 756: What's the reason for reassigning the request parameter?

fredy20501: `req.session` is not a parameter. It is the session object which allow us to store data for a user session. Here we are storing the token so it can be later verified in the create post API. See the npm package [`express-session`](https://www.npmjs.com/package/express-session) for details on how this works.

fredy20501: Short explanation: the session is based on the request since the session id (what tells the backend whose session data to use) is stored in a cookie that is passed with the request by the browser.

nathanielcaron:
It looks like `express-session` docs use 60 seconds for the maxAge of these tokens.
I'm guessing the idea here is to have them last 30 minutes so users can create multiple posts without having to redo the puzzle?

nathanielcaron:
As always, amazing work @fredy20501 🎉 

Everything looks great from a design point of view.

nathanielcaron (approved):

fredy20501:
> It looks like `express-session` docs use 60 seconds for the maxAge of these tokens. I'm guessing the idea here is to have them last 30 minutes so users can create multiple posts without having to redo the puzzle?

Actually, the puzzle is always required when you create a post. The maxAge in this context is the amount of time between creating and completing the puzzle, or the amount of time between completing the puzzle and submitting the post, until the puzzle must be redone.

Example: If you open the puzzle and go do something else for 20 min, when you come back the token is expired so the puzzle needs to be redone (even if you slide it properly it will say it is wrong and give you a new one). Same thing if you complete the puzzle but wait 20min before publishing the post (it will show the error as in the demo and reset the puzzle).

nathanielcaron:
> > It looks like `express-session` docs use 60 seconds for the maxAge of these tokens. I'm guessing the idea here is to have them last 30 minutes so users can create multiple posts without having to redo the puzzle?
> 
> Actually, the puzzle is always required when you create a post. The maxAge in this context is the amount of time between creating and completing the puzzle, or the amount of time between completing the puzzle and submitting the post, until the puzzle must be redone.
> 
> Example (maxAge=5min): If you open the puzzle and go do something else for 5 minutes, when you come back the token is expired so the puzzle needs to be redone (even if you slide it properly it will say it is wrong and give you a new one). Same thing if you complete the puzzle but wait 5 minutes before publishing the post (it will show the error as in the demo and reset the puzzle).

Great, 5 minutes seems like a good compromise then.

sahilsaini432 (approved):
Nice Work Fred 💯

CarterMacLennan (approved):

JoaoPontes98 (approved):

PR #83, opened 2022-02-18
DBO-83,84: Fix various frontend issues

fredy20501:
## Work Completed
- Placeholder avatar now shows up properly in the modal popup.
- Fixed memory leaks for the frontend.
  - The issue was that for loading we use `setTimeout()` to delay actions, but we didn't clear the timers when the component unmounted, leading to an error when the timer stops and the component is not longer mounted. Ex: if you switch pages before loading is done.
- Set a maximum height for the static map in the post so it doesn't stretch too much when the post is long or the screen is small.
- Reduced the max image file size for the frontend (14.5mb) to guarantee it to be smaller than the backend (15mb).
- Increased max body size for requests to the backend & through nginx.
  - Note: I couldn't reproduce the error Nathaniel found so I have no way of knowing if this fixed it.

### DIFF
--- a/Service/app.js
+++ b/Service/app.js
@@ -13,8 +13,8 @@ require('dotenv').config();
 const ENV = process.env;
 
 // API Middleware
-APP.use(EXPRESS.json());
-APP.use(EXPRESS.urlencoded({ extended: true }));
+APP.use(EXPRESS.json({ limit: '2mb' }));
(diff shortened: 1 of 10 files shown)


sahilsaini432 (approved):
Nice Work Fred 💯

nathanielcaron (approved):
👍

CarterMacLennan (approved):

JoaoPontes98 (approved):

PR #85, opened 2022-02-27
Notify users of 429 errors

CarterMacLennan:
Added checks when making calls to the backend for a 429 status code. 

If a 429 status code is received the attatched message is rendered in a react bootstrap alert to the frontend.

Test by updating the router in the backend to have limits in the dev environment, and adjusted the limits to test different cases.

### DIFF
--- a/application/src/api/api.js
+++ b/application/src/api/api.js
@@ -16,10 +16,10 @@ const serviceUrl = ENV.REACT_APP_SERVICE_URL;
 
 // Corresponding 429 Error Message
 const error429Message = [
-  'You have returned to this site too many times this hour. ',
-  'You have created, deleted, or published too many posts this hour.',
-  'You have refreshed user posts too many times this hour.',
-  'You have sent out too many emails this hour.'
+  'You have returned to this site too many times this hour. Please try again later.',
+  'You have created, deleted, or published too many posts this hour. Please try again later.',
+  'You have loaded user posts too many times this hour. Please try again later.',
+  'You have sent out too many emails this hour. Please try again later.'
 ];
 
 // Get Auth token
--- a/application/src/api/api.js
+++ b/application/src/api/api.js
@@ -14,6 +14,14 @@ const MAX_RETRY_LIMIT = 2;
 // Get the service url from the environment file
 const serviceUrl = ENV.REACT_APP_SERVICE_URL;
 
+// Corresponding 429 Error Message
+const error429Message = [
+  'You have returned to this site too many times this hour. ',
+  'You have created, deleted, or published too many posts this hour.',
+  'You have refreshed user posts too many times this hour.',
+  'You have sent out too many emails this hour.'
+];
+
 // Get Auth token
 export const getAuthToken = async () => {
   try {
@@ -28,7 +36,7 @@ export const getAuthToken = async () => {
     if(error?.response?.status === 429) {
       return ({
         error: true,
-        message: error?.response?.data?.message,
+        message: error429Message[error?.response?.data?.errorCode],
       });
     }
   }
@@ -213,7 +221,7 @@ const requestWithToken = async (request) => {
         logger.warn(error);
         return ({
           error: true,
-          message: error?.response?.data?.message,
+          message: error429Message[error?.response?.data?.errorCode],
         });
       }
       if(error?.response?.status === 401) {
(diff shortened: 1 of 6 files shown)


CarterMacLennan:
I believe the last API test failed as something weird happened when I tried to delete the post below. I think if we went into mongo and removed this post it would past this CI test.

`$[9] | actual does not contain expected | actual does not contain key - 'uid' (MAP:MAP)
    {"author":{"name":"dwadwa","avatar_url":"https://***.s3.amazonaws.com/aafcfd249411d8916a27eec035c91b45","__v":0},"body":"dwadwa","tags":["cheap","fall","walking"],"title":"dwadwa","date_created":"2022-02-27T21:10:51.175Z","location":{"lat":45.01534665434719,"lng":-68.16348189468742},"location_string":"Maine, United States","true_location":false,"img_url":"https://***.s3.amazonaws.com/a7fe36de09c3ec5b7967f93aaae89d66","__v":0}`

fredy20501:
I found some issues with the Search & Home pages while testing/reviewing your changes. I fixed them in #86 which is set to merge into this branch. @CarterMacLennan please review that PR and merge it in here if you're ok with it (or leave comments on it if not).

application/src/components/Create.js
```diff
@@ -134,8 +135,10 @@ function Create(props) {
     let imgResult = null;
     if (avatarImg || picture) {
       imgResult = await postImages(avatarImg, picture);
-      if (!imgResult) {
+
+      if (imgResult?.error) {
         setIsCreateError(true);
+        setError429Message(imgResult?.message);
```
fredy20501, line 141: I don't like that we are straight up displaying the error message that is defined in the backend to the users. Those request limit messages are not at all user-friendly. Users might wonder: "What does GET/POST/DELETE mean?", "Why does it mention my IP address?", "What is a 'request'?".

Potential Solutions:
1. Change the message in the backend to be user-friendly.
2. Add a separate message that is meant to be seen by the user, keeping the current one for logging purposes.
3. Instead (or in addition) of a message, return a specific id/number so the frontend can know what limit was reached and define its own messages appropriately.

I prefer option 3 since it keeps the user messages defined within the frontend (reduces coupling with the backend). We can discuss this during our next meeting.

nathanielcaron: I vote for option 3 as well.
The API messages and messages displayed in the application should be different. The API messages are meant for developers or the application using the API.
Using codes would be a clean solution.

e.g.,
response status code: `429`
response body:
```
{
   message: "..."
   errorCode: 1
}
```

CarterMacLennan: @fredy20501 It doesn't look like we can attatch an errorCode to a rate limiter when using: `express-rate-limit`. It seems to only allow you to attatch defined fields.

CarterMacLennan: I'll take another peek, thanks Fred

nathanielcaron: The rate limiter objects already provide a JSON object:
https://github.com/Dubble-bubble-original/engg4000/blob/dev/Service/routes/routes.js#L30

Currently, there is only one property, which is the message. You can add the error code there.

CarterMacLennan: Ahh, shoot I was making the dumb mistake of not doing ".response.data". That was my issue...

application/src/api/api.js
```diff
@@ -14,6 +14,14 @@ const MAX_RETRY_LIMIT = 2;
 // Get the service url from the environment file
 const serviceUrl = ENV.REACT_APP_SERVICE_URL;
 
+// Corresponding 429 Error Message
+const error429Message = [
+  'You have returned to this site too many times this hour. ',
+  'You have created, deleted, or published too many posts this hour.',
+  'You have refreshed user posts too many times this hour.',
+  'You have sent out too many emails this hour.'
+];
```
fredy20501, line 23: It might be good to add the classic "Please try again later." after each of these messages.

For the user post one, I'm not sure if "refreshed" is the right word. Maybe "You have loaded too many posts this hour."? I'm not sure if "loaded" is actually better or not, honestly.

CarterMacLennan: I actually liked using the word loaded so I included it. I thought the "please try again later" was a great idea as well. 

I've just pushed a new commit addressing your comments.

fredy20501 (approved):
🏆

nathanielcaron (approved):
🚀

JoaoPontes98 (approved):

sahilsaini432 (approved):
💯

PR #87, opened 2022-03-03
Fix Geocoding in Production

fredy20501:
## Issue Description
The geocoding API used to get the location as a string based on coordinates cannot be restricted using a domain name, it must use an IP address. To fix this, I changed the geocoding API call to be done by the backend instead of the frontend (since if it is done by the frontend the IP address is always the IP of the user's computer, which we obviously can't control).
## Work Completed
- Changed the geocoding API call to be done by the backend instead of the fronted.
  - There is now a new env variable for the backend for that API key. The new variable is called `GEO_API_KEY` and you should set it to be the same as the `REACT_APP_MAPS_API_KEY` that is defined in the frontend env file (for your dev environment). In production that value is different but I already set it up so don't worry about it.
- Created a new API key specific for production for the geocoding API call. That key is restricted by IP, which is the public ipv4 address of the EC2 instance.
  - Note: the IP will only change if we stop the EC2 instance, in which case I'll just need to update the IP in my Google Cloud Console to match the new IP when the instance is brought back up.


nathanielcaron (approved):
👍

Service/routes/routes.js
```diff
@@ -56,6 +56,14 @@ const emailLimiter = RATE_LIMIT({
   legacyHeaders: false // Disable the `X-RateLimit-*` headers
 });
 
+const geocodeLimiter = RATE_LIMIT({
+  windowMs: 60 * 60 * 1000, // 1 hour
+  max: 100, // Limit each IP to 100 geocode requests per window
```
nathanielcaron, line 61: Is this limit high enough?
If this endpoint is called for every post created and loaded then this seems quite low

Service/routes/routes.js
```diff
@@ -56,6 +56,14 @@ const emailLimiter = RATE_LIMIT({
   legacyHeaders: false // Disable the `X-RateLimit-*` headers
 });
 
+const geocodeLimiter = RATE_LIMIT({
```
sahilsaini432, line 59: Should we use a rate limiter for this endpoint?

fredy20501: Well yeah, we should limit all endpoints so that malicious users can't do a DOS attack on any of our endpoints. Also more specifically, I don't want people to spam this endpoint with thousands of requests cause they aren't free and I'm the one paying 😅

sahilsaini432: That sounds fair

sahilsaini432 (approved):
💯

JoaoPontes98 (approved):

CarterMacLennan (approved):
✅ Great Job!

PR #89, opened 2022-03-07
DBO-70 Bulk Delete

JoaoPontes98:
This PR includes:
- A new endpoint that deletes all users and posts
- A karate feature that uses the end point to bulk delete posts
- Changed the `CI.yml` to use a new DB when running CI tests
- Changed `karate-config.js` to call the bulkDelete feature before running the tests
- Added the bulkdDelete api endpoint to the RAML docs


sahilsaini432 (approved):
💯

Automation/src/test/java/karate-config.js
```diff
@@ -19,5 +19,8 @@ function fn() {
     // customize
   }
 
+  //Clears the DB before performing tests
+  karate.callSingle('classpath:utils/bulkDelete.feature', config);
```
nathanielcaron, line 23: This should only be called if we are running tests in the CI environment, could move it under the else if (env == 'ci') and set the baseUrl like the dev environment.

Could also remove the frontendUrl and the else if (env == 'prod') since we are not using them.

.github/workflows/CI.yml
```diff
@@ -6,7 +6,7 @@ name: CI Pipeline
 # Get .env variables from our repository's secrets
 env:
   NODE_ENV: dev
-  DB_NAME: ${{ secrets.DB_NAME }}
+  DB_NAME: ${{ secrets.DB_CI_NAME }}
```
fredy20501, line 9: This doesn't need to be changed (i.e., we don't need to use a different secret). We just change the **value** of the `DB_NAME` secret variable (in the GitHub secrets) to the database name we want to use.

Automation/src/test/java/karate-config.js
fredy20501: In that case, we also need to change the `NODE_ENV` in the `CI.yml` file to be "ci":
```yml
NODE_ENV: ci
```
We also need to change everywhere else that we check for `NODE_ENV === 'dev'` to also include 'ci'. I think there are only 2 other places: line 43 in `app.js` and line 60 in `routes.js`. We can simply change those to be like this:
```js
if (ENV.NODE_ENV === 'dev' || ENV.NODE_ENV === 'ci')
```

CarterMacLennan (approved):

nathanielcaron (approved):

PR #91, opened 2022-03-09
DBO-64, DBO-78 - Failing Karate UI Command and Cron Job for Running Tests

nathanielcaron:
- Fixed failing UI automation tests
- Updated runner to only use a single thread to fix the failing maven command
- Added E2E.yml file to run UI automation tests as a cron job daily


.github/workflows/E2E.yml
```diff
@@ -0,0 +1,39 @@
+name: E2E Tests
```
fredy20501, line 1: I don't think running the tests on the production website is a good idea. The point of tests is to find issues **before** they are pushed to production. Testing should ideally happen before merging into `dev`, or at the very least before merging into `master` (i.e., before deployment). Also if the test for deleting a post fails, then we will end up leaving "garbage" posts on the production website.

Ideally, we could have a dev deployment that we can use to run the tests against so we're not polluting the production website if they fail and we are at least testing before merging into master.

I think the best we can do with what we have currently is to just make sure we always run the UI tests manually in our dev environment on each PR before merging them.

fredy20501 (approved):

CarterMacLennan (approved):

sahilsaini432 (approved):
💯

JoaoPontes98 (approved):

PR #94, opened 2022-03-10
DBO 36 - Create Karate UI automation tests to search for posts in search page

sahilsaini432:


application/src/components/Search.js
```diff
@@ -164,13 +166,14 @@ function Search() {
               <FRow>
                 <FCol>Tags:</FCol>
                 <FCol className="w-100">
-                  <div className="tag-container">
+                  <div className="tag-container" data-testid="selcted-tags">
```
fredy20501, line 169: Typo: `selected-tags`

Automation/src/test/java/ui/searchpage.feature
```diff
@@ -0,0 +1,91 @@
...
+
+    # Enter search page
+    When click('a[data-testid=nav-search-btn]')
+    Then waitFor('div[data-testid=filter-input]')
+
+    # Search for a tag
+    When input('#filter-input-field', 'Se')
+    Then click('#tgb-btn-desert')
+    And click('#tgb-btn-seasonal')
+    And match enabled('button[data-testid=search-button]') == true
```
fredy20501, line 85: For this filter test, we should check that other tags are no longer on the page (i.e., test that the filter actually hides the other tags). We can probably use the `exists()` method and test for a few tags that we know don't match the filter.

Automation/src/test/java/ui/searchpage.feature
```diff
@@ -0,0 +1,91 @@
...
+  Scenario: Enter Search page
+    # Enter website
+    Given driver frontendUrl
+    When click('input[data-testid=agree-checkbox]')
+    And click('button[data-testid=enter-btn]')
+    Then waitFor('nav[data-testid=navbar]')
+
+    # MOve to search page
+    When click('a[data-testid=nav-search-btn]')
+    Then waitFor('div[data-testid=search-page]')
```
fredy20501, line 16: I don't think we need this scenario since all the other scenarios are also doing this, so we'll know if it fails even without this one.

Automation/src/test/java/ui/searchpage.feature
```diff
@@ -0,0 +1,91 @@
...
+    And click('button[data-testid=enter-btn]')
+    Then waitFor('nav[data-testid=navbar]')
+
+    # Enter search page
+    When click('a[data-testid=nav-search-btn]')
+    Then waitFor('div[data-testid=tag-group]')
+
+    # Select a tag
+    When click('#tgb-btn-fall')
+    Then match enabled('button[data-testid=search-button]') == true
```
fredy20501, line 44: I don't think we need this scenario since the one just below ("Click search button") is also doing the exact same thing at the beginning.

fredy20501 (approved):

nathanielcaron (approved):

CarterMacLennan (approved):

JoaoPontes98 (approved):

PR #95, opened 2022-03-14
DBO-92: Add Red Exclamation mark icon to post with tooltip describing why a post is flagged

sahilsaini432:
**Changes Made**

https://user-images.githubusercontent.com/32021832/158253744-3ed40c7b-ce07-4922-85d5-c4a5b6598f63.mov


application/src/components/post/post.css
```diff
@@ -10,19 +10,76 @@
...
+    background: #fff;
+    box-shadow: 0 10px 10px rgba(0,0,0,0.3);
+    padding: 10px 18px;
+    border-radius: 25px;
+    opacity: 0;
+    pointer-events: none;
+    transition: all 0.4s cubic-bezier(0.68,-0.55,0.265,1.55);
+}
+
+.aleat-symbol {
```
fredy20501, line 55: Typo: `.alert-symbol`

application/src/components/post/post.css
```diff
@@ -10,19 +10,76 @@
...
+}
+
+.custom-tooltip {
+    color: white;
+}
+
+.flagged .custom-tooltip {
+    position: absolute;
+    left: 0px;
+    width: 900%;
```
fredy20501, line 45: For the width, I think it would be better to use `max-content` so it actually takes the exact width of the text.
```css
width: max-content;
```
| Before | After |
| ------ | ----- |
| image| image |

application/src/components/post/post.css
```diff
@@ -109,6 +166,11 @@
     .avatar-column {
         width: 70px;
     }
+
+    .flagged .custom-tooltip {
+        /* Reduce tooltip text size */
+        width: 500%;
+    }
```
fredy20501, line 173: Using a % unit doesn't make sense here. We're basically sizing the pop-up to 5 times the width of the icon, but the popup size shouldn't be related to the size of the icon. I think a fixed pixel size would make more sense.
```css
width: 250px;
```

fredy20501 (approved):
Nice work! 🥇 🔥

CarterMacLennan (approved):

JoaoPontes98 (approved):

PR #97, opened 2022-03-18
DBO-76: Add React-Router to frontend

CarterMacLennan:
## Work Completed
- Added React-Router to the frontend.
  - Used HashRouter to keep the URL the same on page refresh.
- Added global state to keep track of whether user has accepted the terms of service.
  - This is stored in the browser's local storage, which is permanent.
  - If the user has not accepted the terms, all URLs will redirect to the welcome page.

## Note
If you want to see/clear the persistent state, it is under Application > LocalStorage in the browser's devtools.
image


fredy20501 (approved):

nathanielcaron (approved):

sahilsaini432 (approved):
Nice Work 💯

JoaoPontes98 (approved):

PR #99, opened 2022-03-26
Investigate and fix CI/CD Pipeline

sahilsaini432:
**Changes Made**

- Updated verifyImage endpoint to not save images anymore.
- Updated service docker image to be `buster-slim` instead of `alpine`
- Updated service docker file to use `ecosystem.config.js` file to start the service
- Updated server to use 2600MB memory
- Updated postimage endpoint to compress images before uploading to the S3 bucket. (now every image is saved as a jpeg).
- Removed https://github.com/Dubble-bubble-original/engg4000/blob/dev/Automation/src/test/java/api/image.feature#L71-L76 Automated API test, this won't pass anymore because the images are compressed.


nginx/configs/default.conf
```diff
@@ -16,12 +16,12 @@ server {
 
     location / {
         proxy_pass http://client/;
-        client_max_body_size 2M;
+        client_max_body_size 1G;
```
nathanielcaron, line 19: This setting is for the request body size:
http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size

1G seems outrageously high 😄

sahilsaini432: Ya, I was just trying something out 😄

nathanielcaron (approved):

nathanielcaron:
We did it! 🥳 
Great work @sahilsaini432

fredy20501 (approved):
Woo! 🥳 🔥 🎆

JoaoPontes98 (approved):
You the man!

CarterMacLennan (approved):

PR #100, opened 2022-03-28
 DBO-98 - (BUG) Investigate why image uploads fail

nathanielcaron:
- Updated max image upload size to be 5MB
- Updated max request size to be 12MB (Enough for two 5MB images and some buffer)
- Added loading spinner when post is being published so users can hit submit several times

Why did I do this?
When I increased the max request size to be 30MB to accommodate image uploads of 15MB, it exposed a couple of issues. First, uploading these large images takes a really long time (> 30 seconds). Second, the uploads were taking so long that the endpoint was returning before it had received the image key.
From my testing, having the limit set at 5MB works consistently and doesn't lead to any issues.


fredy20501 (approved):
Awesome work! 💯

CarterMacLennan (approved):
Great catch, PR looks great 2 me

fredy20501 (commented):
Since it still takes a while (a few seconds) to upload the images we definitely need to have some feedback for the user during that time. At the very least we should disable the publish button (otherwise you can publish the same post a bunch of times - I just did exactly that), but ideally we would have a loading spinner too.

nathanielcaron:
> 

Agreed

nathanielcaron:
> Since it still takes a while (a few seconds) to upload the images we definitely need to have some feedback for the user during that time. At the very least we should disable the publish button (otherwise you could publish the same post a bunch of times), but ideally we would have a loading spinner too.

Agreed

fredy20501 (approved):
Nice loading spinner 💯

sahilsaini432 (approved):
Looks good to me 💯

